PR #106, opened 2010-10-14
Truly cancel timers

phs:
See http://github.com/eventmachine/eventmachine/issues/#issue/89

This patch adds EventMachine_t::UninstallOneshotTimer, which removes a timer from the internal multimap.  A new map is added, taking signatures to firing times, to keep this removal efficient.

The method is exposed to ruby as EM.cancel_oneshot_timer via evma_uninstall_oneshot_timer.  EM.cancel_timer now delegates to it, instead of setting the timer callback to false and ignoring it later.

A timer may still safely cancel itself (a no-op) from within its own callback.

The pure-ruby reactor (which relied on the logic in EM.cancel_timer) also supports #cancel_oneshot_timer.  emwin.cpp, cplusplus.cpp and the java port have not been touched.

### DIFF
--- a/ext/cmain.cpp
+++ b/ext/cmain.cpp
@@ -107,10 +107,10 @@ extern "C" const unsigned long evma_install_oneshot_timer (int seconds)
 evma_uninstall_oneshot_timer
 **************************/
 
-extern "C" int evma_uninstall_oneshot_timer (const unsigned long binding)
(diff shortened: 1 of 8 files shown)


sodabrew:
This looks very interesting, but has been sitting a long time and will need some updating to merge. On my radar now.

PR #114, opened 2010-10-27
Restartable timer

adzap:
Added a restartable timer class. It's a one-shot timer but can be restarted from the beginning of the interval if restarted before it is fired or cancelled.

I use it in a project so that I can restart the timer on a user interrupt event without needing to cancel and recreate a timer with the same interval.

I haven't done the java class for it yet as I just wanted to see if you're interested.

Adam


sodabrew:
Since this is included in Eventmachine-LE, I'm tagging it for v1.2.0 with #570

PR #165, opened 2011-02-06
Support seamless upgrades of EM servers

pc:
The two commits below allow you to have a server running, and then, fork(), exec(), start a new server using the same socket, finish up requests in the master, and then start processing new requests in the child. This allows you to do upgrades without dropping any requests.


sodabrew:
Can this be closed since #465 landed a while ago?

PR #196, opened 2011-05-03
Added support for the NPN TLS when it is available.

carsonmcdonald:
Instructions for building against a version of OpenSSL that supports NPN can be found here along with an example for use with SPDY: https://gist.github.com/944386

### DIFF
--- a/ext/cmain.cpp
+++ b/ext/cmain.cpp
@@ -431,6 +431,32 @@ extern "C" void evma_set_tls_parms (const unsigned long binding, const char *pri
 		ed->SetTlsParms (privatekey_filename, certchain_filename, (verify_peer == 1 ? true : false));
 }
 
+/****************************
+evma_set_negotiable_protocols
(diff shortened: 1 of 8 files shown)


SkyEyes:
HTTP-2 support) will be more needed at this moment :) but with such speed we can wait for HTTP-2 when HTTP-4 will be released :)

sodabrew:
@SkyEyes I'm not sure I understand your sentiment, but I do urge you to take a look at the recent updates to EventMachine - it is once again being actively maintained. I do plan to dig into NPN / ALPN support for the EM 1.3 release. Based on my own availability, I'd say it's 6-12 months out, but Pull Requests are very much welcome if you'd like to help make it happen sooner!

SkyEyes:
I mean nothing against maintainers and their speed at all :)
Need involve more people for this feature :) and we will get it much faster for sure :)

PR #202, opened 2011-05-13
Help Rubinius out with collecting garbage

evanphx:
This patch adds a one_shot_only method which has run_machine not run a while(true) loop forever. Rubinius uses this mode so that run_machine can return and thus allow all objects accessed within run_machine to be collected as garbage. Because of the way ruby allows C extensions to work, Rubinius must keep a strong reference to all objects referenced by a C function. run_machine looping forever causes a memory leak and thus doing the loop in ruby allows those things to be freed.


sodabrew:
@evanphx @brixen Do you know if this is still a problem with current Rubinius?

evanphx:
@sodabrew Yeah, this is still a problem. (wow this PR is 3.5 years old and the first comment is an hour ago).

sodabrew:
So the issue is that you need the extension to "come up for air" once in a while, back into Ruby space, so that you can garbage collect?

I wonder if we could always pump the machine this way, or if it would kill MRI performance (I suspect it would, but...).

evanphx:
@sodabrew Yes, that's correct. I suspect it wouldn't hurt performance on MRI either, given that most of the time is spent waiting on the IO events anyway.

I'll see about doing a rebase.

evanphx:
@sodabrew There ya go, applies to master now.

sodabrew:
Resolved by #617

PR #235, opened 2011-08-08
Update license notifications

nahi:
Follow-up for the documentation revamp at 38705c44.
- LICENSE -> COPYING: Source files are still pointing COPYING file
- GNU -> GPL: Follows the file name in CRuby distribution
- LEGAL: Recovered from docs/old since it has lib/em/buftok.rb.

lib/em/buftok.rb is covered with exactly the same term as EventMachine
itself so LEGAL file would not be needed (I'm not a lawer!)


CyborgMaster:
The official Ruby License has now changed quite a bit.  It is now BSD based instead of GPL, so some people (myself included) can find the reference to the "Ruby License" confusing.  

Are you still considering changing to MIT?  That would make it very obvious what is and is not allowed.  

If not, can I still use this gem as a dependency for my proprietary, closed source application without having to release it as open source?

PR #241, opened 2011-08-25
Allow compilation of EM on MagLev

timfel:
MagLev sports C extension support now, but it's far from complete. These changes get the gem to compile and load on MagLev.


sodabrew:
I suspect this is no longer needed (?)

PR #290, opened 2012-01-15
Test Unit for IPv6

ibc:
Test unit for testing IPv6 under TCP and UDP. The test unit checks both EM client and EM server.

For the test unit to work, a local IPv6 is required in the computer where the test runs. In Linux I use Miredo (which also runs in Windows and MacOSX): http://www.remlab.net/miredo


sodabrew:
Closing in favor of PR #570

PR #303, opened 2012-03-06
Added windows IPv6 support

stakach:
Also increased file and socket limits


sodabrew:
Note to self: Check if #570 includes this work.

sodabrew:
Superseded by #630

PR #306, opened 2012-03-13
Making outbound data queue size available to user

rlevin:
Hello,

When using eventmachine as a TCP server, it can be useful to know when the network buffers have filled up and we are starting to queue inside of the EventableSocketChannel (so you can take action and close the socket as necessary, to avoid overflowing the heap and crashing your process).  The C ruby version has a way of accessing this internal queue, so I've added it to the jruby version.

Thanks :)
--becky


sodabrew:
Superseded by #556 and available in eventmachine 1.0.6 and newer.

PR #317, opened 2012-04-19
Add possibility to watch on outbound data

funny-falcon:
EM::Connection produced by EM.connect has no ability to react on slow clients, `send_data` pushes data infinitely.

This pull request two ability:
1. to know outbound data size . It is already exists on C level, but were not exposed to Ruby level

```
  c = EM.connect('xx','xx', EM::Connection)
  c.send_data('a'*1000000)
  c.outbound_data_size
```
1. to react on outbound data size reduction

```
  module Handler
    def sent_data
      puts "some data sent"
    end
  end
  c = EM.connect('xx','xx', EM::Connection){|c| c.notify_sent_data = true}
  c.send_data('a'*1000000)
```

### DIFF
--- a/java/src/com/rubyeventmachine/EmReactor.java
+++ b/java/src/com/rubyeventmachine/EmReactor.java
@@ -49,6 +49,7 @@ public class EmReactor {
 	public final int EM_SSL_VERIFY = 109;
 	public final int EM_PROXY_TARGET_UNBOUND = 110;
     public final int EM_PROXY_COMPLETED = 111;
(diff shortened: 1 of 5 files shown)


funny-falcon:
I've added jruby implementation + added implementation for 'pause/resume/get_outbound_data_size'.
Removed first commit - 'get_outbound_data_count' - although i still think it is usefull.

funny-falcon:
Looks like travis fails randomly on other tests. I can run `test_sent_data.rb` without issue.

PR #323, opened 2012-04-27
store path inside of FileWatch.new

funny-falcon:
Store path before constructor called, so that we are not forced to pass it twice for custom class


sodabrew:
Would this change still be valuable to merge? If so, could you rebase? Thanks!

PR #342, opened 2012-07-01
Extract from buffer in edge case of multi-character delimiter

johnkchow:
The BufferedTokenizer works great with single character delimiters, but with multi-character delimiter it doesn't address the edge case of extracting the delimiter out of the internal buffer. The original discussion as well as code examples of the edge case can be found in another pull request:

https://github.com/eventmachine/eventmachine/pull/338


sodabrew:
Resolved by #547

PR #397, opened 2013-01-27
New :cert_auth_file option for EventMachine::Connection#start_tls

rampion:
Full disclosure: this was inspired by a [question I asked on StackOverflow](http://stackoverflow.com/questions/14470042/how-do-i-choose-a-certificate-authority-file-when-using-thin).

This commit adds the `:cert_auth_file` option to `EventMachine::Connection#start_tls` so that users can have the framework preverify client certificates using a list of approved certificate authorities.  The preverification result will be passed as a second argument to `#ssl_verify_peer`.  The default `#ssl_verify_peer` will rely upon the preverification result to determine its return value.


parhs:
why is this not merged?

PR #420, opened 2013-03-12
Add support for EM.set_simultaneous_accept_count(<int>).

macournoyer:
This controls the number of calls to `accept(2)` done in one turn in the loop. The default value is `10`, like before, so it won't change the default behaviour of EM.

The purpose of this it to prevent EM from accepting more connections than it can handle. 
My end goal is to set this to `1` in Thin.

Also add EM.get_simultaneous_accept_count to retreive the value.

If necessary, I can port my changes to the Java backend. Let me know.

/cc @raggi

### DIFF
--- a/ext/cmain.cpp
+++ b/ext/cmain.cpp
@@ -642,7 +642,6 @@ extern "C" int evma_get_max_timer_count()
 	return EventMachine_t::GetMaxTimerCount();
 }
 
-
 /************************
 evma_set_max_timer_count
 ************************/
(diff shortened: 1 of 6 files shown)


sodabrew:
@macournoyer Are you still pursuing Thin 2, and would this be of value even to Thin 1?

sodabrew:
Ok! Actually I was going to suggest maybe calling this `EM.accept_per_crank` - and begin linking up a group of `*crank` options. Like the one for Rubinius cranking in Ruby instead of inside the `EventMachine->Run()` method.

raggi:
I proposed EM.crank() for single-step EM a long time ago (useful for tests, as you can also then do crank_until(timeout), etc). @tmm1 didn't like it.

I'm not really sure what this commit is expecting to change though, in terms of real world semantics. Modern OSes in default configurations will pre-accept anyway, and app pulling another 20 off a queue of 1024 (linux) won't make a huge difference in terms of balancing, etc. I could see this helping reduce dead letter reads by some proportional amount, and maybe it produces just enough backpressure to improve some semantics, but overall we'd need a lot more tunables to fix accept load balancing.

PR #438, opened 2013-04-25
Add a start_tls option to specify the minimum SSL version required

fabiokung:
These patches allow EM code to disable SSLv2, or force TLSv1 only.

Usage:

```ruby
module Handler
  def post_init
    start_tls(:min_version => :sslv3)
  end
end
```

`:min_version` is case insensitive and can be `:sslv2` (any, default), `:sslv3` (disables SSLv2), or `:tlsv1` (TLSv1 only).

This is backwards compatible and the default is the same as before (`:min_version => :SSLv2`). @tmaher can fill us in with reasons why someone would want to completely disable SSLv2 and/or SSLv3 nowadays.


sodabrew:
I don't think a min version is the right way to go. It's entirely possible that a bug would be discovered in a "middle version" -- for example, TLSv1.1 could become compromised while TLSv1 and TLSv1.2 are safe. We'll need a way to handle that scenario.

jonbrenner:
I agree with @sodabrew. In addition to specifying specific versions due to security concerns, there are interoperability reasons for specifying specific versions. E.g., embedded systems with poor implementations of specific protocol versions.

sodabrew:
This is cherry-picked and further developed in #654

PR #452, opened 2013-07-23
prevent busy loop in ruby 1.8.7 exception handling

manlon:
Not sure what the support policy is for 1.8.7, but I ran into a case where the EM thread enters a busy loop and uses ever-increasing memory if EM.run happens to be called from an exception handler. Here is a minimal repro:

https://gist.github.com/manlon/6058884

The behavior or $! seems fixed in ruby 1.9 but it is easy enough to work around here.

### DIFF
--- a/lib/eventmachine.rb
+++ b/lib/eventmachine.rb
@@ -957,13 +957,16 @@ module EventMachine
       callback = @next_tick_mutex.synchronize { @next_tick_queue.shift }
       begin
         callback.call
+      rescue
+        exception_raised = true
+        raise
       ensure
         # This is a little nasty. The problem is, if an exception occurs during
         # the callback, then we need to send a signal to the reactor to actually
         # do some work during the next_tick. The only mechanism we have from the
         # ruby side is next_tick itself, although ideally, we'd just drop a byte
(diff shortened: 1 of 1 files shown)


sodabrew:
I cannot reproduce this.

sodabrew:
Repro'd! Code based on the gist from #482, made even smaller and with a bonus to really see the output...

```ruby
require 'eventmachine'

begin
  raise
rescue Exception
  EM.run do
    EM.next_tick do
      EM.add_periodic_timer(0.1) { print '.' }
      EM.add_periodic_timer(1) { puts '.' }
    end
  end
end
```

Without the patch, the first line has 8-9 dots. The second line has 8ish, the third line 7ish, the fourth line 6ish, and so on. CPU fans rev up. The process is clearly in a busy loop.

With the patch, the first line and all subsequent lines have 9-10 dots. CPU fans stay calm.


I was having trouble with the repro because I wasn't using `EM.next_tick` inside the `EM.run` block. That seems to be crucial to the failure; if you put the calls to `EM.add_periodic_timer` directly inside `EM.run` then you'll always get the 9-10 dots, calm CPU behavior.

PR #453, opened 2013-07-23
Added EM::Connection#close_write

genki:
EM.popen needs to half-close the duplex descriptor so that the subprocess can detect EOF.

Here is typical use case:

``` ruby
EM.run do
  EM.popen "cat", (Module.new do
    define_method :post_init do
      send_data "hello\n"
      EM.next_tick{close_write}
    end
    define_method :receive_data do |data|
      puts data
    end
    define_method(:unbind){EM.stop}
  end)
end
```


sodabrew:
I don't see a close_write method in the current state of this branch, are you missing some commits?

PR #271, opened 2011-10-10
Add EventMachine::attach_server

ramonmaruko:
EventMachine::attach_Server is for attaching an arbitrary server IO
object or file descriptor to the event loop.

Tries to fix #93.

Signed-off-by: Ramon Marco L. Navarro ramonmaruko@gmail.com

PR #465, opened 2013-09-25
Add EventMachine.attach_server

benburkert:
This adds the `EventMachine.attach_server` & `EventMachine.attach_socket_server` methods for registering an existing FD/socket created outside of EventMachine. It's a simplified version of #271, but without the CLOEXEC/autoclose stuff. This allows sockets to be shared between parent & child processes like with unicorn's fork-then-exec graceful restart functionality.


sodabrew:
Was this ever implemented for JRuby? https://github.com/eventmachine/eventmachine/pull/271/files#discussion_r22436804

If not, the test should be skipped on JRuby.

PR #486, opened 2014-01-08
Handle deferred exceptions in reactor thread

joshua-rutherford:
In the current implementation, if a deferred thread raises an exception while the main thread is executing a very indiscriminate rescue block (e.g., Logger::LogDevice.write) the error may be handled improperly.  In the case of the LogDevice, it will get swallowed up entirely never to be seen again.  This is a bad thing.


sodabrew:
Interesting idea. Sidenote, the second commit, "SQUASH" has only a whitespace change - squash it away?

joshua-rutherford:
@sodabrew, I've removed that commit, rebased on master and corrected a Travis failure on 1.8.7.

sodabrew:
This looks great. The change in arguments to EM.defer appears to create an API change - for that reason I think I should schedule this for 1.2.0. OK with you?

joshua-rutherford:
Sure.  Thanks.

PR #488, opened 2014-01-16
Create file descriptors with O_CLOEXEC where possible

ebroder:
Where possible and supported by the operating system, pass creation-time flag arguments that create close-on-exec file descriptors. On platforms (e.g. OS X) that don't support pipe2(2), accept4(2), or SOCK_CLOEXEC, fall back to creating non-close-on-exec descriptors.

Based on patches by Greg Brockman, Patrick Collison

### DIFF
--- a/ext/ed.h
+++ b/ext/ed.h
@@ -27,7 +27,7 @@ class SslBox_t; // forward reference
 #endif
 
 bool SetSocketNonblocking (SOCKET);
-
+bool SetFdCloexec (int);
 
 /*************************
 class EventableDescriptor
@@ -418,5 +418,3 @@ class InotifyDescriptor: public EventableDescriptor
 };
 
 #endif // __EventableDescriptor__H_
-
-
--- a/ext/ed.cpp
+++ b/ext/ed.cpp
@@ -31,7 +31,7 @@ bool SetSocketNonblocking (SOCKET sd)
 	int val = fcntl (sd, F_GETFL, 0);
 	return (fcntl (sd, F_SETFL, val | O_NONBLOCK) != SOCKET_ERROR) ? true : false;
 	#endif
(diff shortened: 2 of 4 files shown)


sodabrew:
I merged #591 this week, for the 1.0.8 release. Could you rebase this PR and take into consideration the notes from https://github.com/eventmachine/eventmachine/pull/591#issuecomment-125338450 (use the extra args if available at compile time, fall back to open-then-set if not available at runtime or compile-time, for maximum best-effort).

antifuchs:
Hi, I've taken on forward-porting this (thanks to @ebroder for giving me access!). This change should take care of `fcntl`ing when setting CLOEXEC on open fails or isn't available, and in the process made #591's change use the same mechanism.

Hope that's acceptable; as always, let me know if anything needs attention (:

ext/ed.h
```diff
@@ -27,7 +27,7 @@ class SslBox_t; // forward reference
 #endif
 
 bool SetSocketNonblocking (SOCKET);
-
+bool SetFdCloexec (int);
```
sodabrew: Can be `SOCKET`, there's a `typedef int SOCKET` in project.h (for consistency with `SetSocketNonblocking`).
Name it `SetSocketCloexec` for consistency as well?

sodabrew:
This is awesome.

ext/ed.h
antifuchs: Yeah, I was reluctant to do that since we apply it to non-socket FDs, too. Happy to perform the rename if you say that's ok (-:

sodabrew: Oh, duh - nonblocking applies only to socket, cloexec applies to any FD. Let's keep it as-is.

PR #490, opened 2014-01-16
Add some basic tests that the DNS resolver isn't leaking timers

ebroder:
It's a pretty crude test - among other things, it relies on `stop_timer` only setting `@timer` to nil once it's cancelled the timer.

But it's hard to do better - since `@timer` does in fact get nil'd out, we can't poke to see if the timer was cancelled, and since cancellation of periodic timers doesn't deregister the callback, we can't look at `EM.instance_variable_get(:@timers)`.

### DIFF
--- a/tests/test_resolver.rb
+++ b/tests/test_resolver.rb
@@ -52,4 +52,30 @@ class TestResolver < Test::Unit::TestCase
       }
     }
   end
+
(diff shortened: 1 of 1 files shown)


sodabrew:
Merged at #571

PR #502, opened 2014-03-12
Use the rb_fd_* API to get autosized fd_sets

### LINKED ISSUE #501: EM select loop crashes when fd >= FD_SETSIZE
The [_RunSelectOnce](void EventMachine_t::_RunSelectOnce()) function crashes the whole Ruby VM when it hits FDs over FD_SETSIZE.  In my case, a too-high fd becomes available for reading.  The `fdreads` and `fdwrites` sets overlap, so when fd *N* is available for reading, EM thinks that fd *N-1024* is available for writing.  The descriptor in question is a server socket (AcceptorDescriptor), which `throw`s in its [Write function](https://github.com/eventmachine/eventmachine/blob/master/ext/ed.cpp#L1477-L1481).  Down comes the VM.

It appears I can fix this with a simple `EM.epoll`, so it's not urgent for me.  But it would be nice to have `assert(fd < FD_SETSIZE)` before each call to `FD_SET` [here](https://github.com/eventmachine/eventmachine/blob/master/ext/em.cpp#L891-L901).  Then at least the crashes would be straightforward and in the right place.

It's also possible, at least on Linux, to `malloc` `fd_set`s with more than `FD_SETSIZE` bits in them.  So instead of crashing when `fd >= FD_SETSIZE`, just `realloc`.

I'm happy to write and PR either solution.

piki:
Fixes #501

On Ruby 1.9.1 and up, we use `rb_fd_select` and friends to get automatically resized `fd_set`s.  This should transparently provide correct behavior for `select` even when fds exceed `FD_SETSIZE` (1024).

On Ruby 1.8.7 and 1.9.0, we use stubs that warn to stdout whenever anyone tries to `FD_SET`, `FD_CLR`, `FD_ISSET`, or `select` an fd >= `FD_SETSIZE`.  This behavior isn't correct, in that it's noisy and doesn't do anything useful with the high FDs.  But it's better than the current behavior, which is undefined and usually crashes with an unrelated assert or heap corruption after the first high FD.

There's another branch [here](https://github.com/piki/eventmachine/blob/backport-rb-fd/ext/rbselect.impl) that backports the rb_fd_* API.  I've only tested that branch on Debian squeeze (6.0), which has Ruby 1.8.7.  It's unlikely to work on WIN32 without further effort.  I can merge it into this PR if someone wants to tidy it up for Ruby <= 1.9.0 on WIN32.

### DIFF
--- a/ext/em.h
+++ b/ext/em.h
@@ -66,6 +66,9 @@ See the file COPYING for complete licensing information.
 
 #ifndef rb_fd_max
 #define fd_check(n) (((n) < FD_SETSIZE) ? 1 : 0*fprintf(stderr, "fd %d too large for select\n", (n)))
+// These definitions are cribbed from include/ruby/intern.h in Ruby 1.9.3,
+// with this change: any macros that read or write the nth element of an
+// fdset first call fd_check to make sure n is in bounds.
 typedef fd_set rb_fdset_t;
 #define rb_fd_zero(f) FD_ZERO(f)
 #define rb_fd_set(n, f) do { if (fd_check(n)) FD_SET((n), (f)); } while(0)
--- a/ext/em.h
+++ b/ext/em.h
@@ -64,6 +64,27 @@ See the file COPYING for complete licensing information.
   #define EmSelect rb_fd_select
 #endif
 
+#ifndef rb_fd_max
+#define fd_check(n) (((n) < FD_SETSIZE) ? 1 : 0*fprintf(stderr, "fd %d too large for select\n", (n)))
+typedef fd_set rb_fdset_t;
+#define rb_fd_zero(f) FD_ZERO(f)
+#define rb_fd_set(n, f) do { if (fd_check(n)) FD_SET((n), (f)); } while(0)
+#define rb_fd_clr(n, f) do { if (fd_check(n)) FD_CLR((n), (f)); } while(0)
+#define rb_fd_isset(n, f) (fd_check(n) ? FD_ISSET((n), (f)) : 0)
+#define rb_fd_copy(d, s, n) (*(d) = *(s))
+#define rb_fd_dup(d, s) (*(d) = *(s))
+#define rb_fd_resize(n, f)  ((void)(f))
+#define rb_fd_ptr(f)  (f)
+#define rb_fd_init(f) FD_ZERO(f)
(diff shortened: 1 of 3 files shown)


ext/em.h
```diff
@@ -66,7 +61,28 @@ See the file COPYING for complete licensing information.
...
+#define rb_fd_resize(n, f)  ((void)(f))
+#define rb_fd_ptr(f)  (f)
+#define rb_fd_init(f) FD_ZERO(f)
+#define rb_fd_init_copy(d, s) (*(d) = *(s))
+#define rb_fd_term(f) ((void)(f))
+#define rb_fd_max(f)  FD_SETSIZE
+#define rb_fd_select(n, rfds, wfds, efds, timeout)  \
+  select(fd_check((n)-1) ? (n) : FD_SETSIZE, (rfds), (wfds), (efds), (timeout))
+#define rb_thread_fd_select(n, rfds, wfds, efds, timeout)  \
+  rb_thread_select(fd_check((n)-1) ? (n) : FD_SETSIZE, (rfds), (wfds), (efds), (timeout))
```
sodabrew: Are these definitions copied from somewhere or newly written here?

sodabrew:
Is this still useful? If so, please rebase to eliminate the extra merge commit.

ext/em.h
piki: Done.

piki:
> please rebase

Done and force-pushed.

sodabrew:
Awesome, thanks! I'll get this on my PR review queue.

sodabrew:
Could you rebase to master once more - unit tests are now running well on Travis!

piki:
> Could you rebase to master once more

Done.

ext/em.h
tmm1: Leaving a quick note since I was initially confused why these were here. The `ifndef rb_fd_max` guard on this section means these macros are only used with really old rubies which did not include rb_fd_* in ruby/intern.h. This provides a simple compat layer around FD_SETSIZE and we don't need to worry about keeping it up to date with changes in ruby itself.

tmm1:
This PR introduced a memory leak in v1.0.5 and v1.0.6 (https://github.com/eventmachine/eventmachine/issues/585#issuecomment-73646276) because `rb_fd_init` allocates memory on the heap.

PR #519, opened 2014-06-20
Add Ruby 2.0.0 and Ruby 2.1.2 to Travis build matrix

greysteil:
Would be great to know that Ruby 2.1.2 is supported.


sodabrew:
Thank you! But this is now superseded by #550 and #551.

PR #523, opened 2014-07-26
"Helvetica,Arial,FreeSans"

IvanUkhov:
Hello,

The value of the font-family property is potentially not what was intended: when wrapped in quotation marks, it is treaded as a single typeface name, not as a list of alternatives.

Regards,
Ivan


sodabrew:
Good fix, thank you.

Issue #524, opened 2014-07-31
Bug: Assertion failed: (!bWatchOnly), function Write, file ed.cpp, line 927.


jim-slattery-rs:
The following script demonstrates a bug in `EventMachine::_RunSelectOnce()`, producing [this assertion failure](https://github.com/eventmachine/eventmachine/blob/v1.0.3/ext/ed.cpp#L927):

```ruby
require 'eventmachine'

module SimpleClient
  def notify_writable
    puts "Being bad in callback: setting $conn2.notify_writable = false"
    $conn2.notify_writable = false
  end
end

EM.run do
  r1, w1 = IO.pipe
  r2, w2 = IO.pipe

  puts "Adding EM.watches"
  $conn1 = EM.watch(r1, SimpleClient)
  $conn2 = EM.watch(r2, SimpleClient)

  $conn1.notify_writable = true
  $conn2.notify_writable = true

  puts "stopping EM"
  EM.stop
end
```

When run again EM v1.0.3 or earlier, produces this output:
```
$ bundle exec ruby create_em_crash.rb
Adding EM.watches
stopping EM
Being bad in callback: setting $conn2.notify_writable = false
Assertion failed: (!bWatchOnly), function Write, file ed.cpp, line 927.
Abort trap: 6
```

Here is the assertion that fails: https://github.com/eventmachine/eventmachine/blob/v1.0.3/ext/ed.cpp#L927

This script takes advantage of the fact that `EM::_RunSelectOnce` iterates over the connections twice: once to determine which ones to call `Write()` on, and then once to call `Write()` on each of them.

But there's room for error, as state may change in the meantime before `Write()` is finally called.  And that is what I exploit to get `Write()` to be called when `bWatchOnly` is true, and `bNotifyWritable` is false.

I have a fix in mind (call `SelectForWrite()` once more, before calling `Write()`), and will create a PR.

sodabrew:
Resolved by #525 thank you!

PR #525, opened 2014-08-01
Avoid calling Write() if SelectForWrite() value has changed

jim-slattery-rs:
This fixes bug #524.

Added a new unit test which fails in current version of EM, but passes with the included fix to em.cpp.

### DIFF
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -973,8 +973,12 @@ void EventMachine_t::_RunSelectOnce()
 					continue;
 				assert (sd != INVALID_SOCKET);
 
-				if (rb_fd_isset (sd, &(SelectData.fdwrites)))
-					ed->Write();
+				if (rb_fd_isset (sd, &(SelectData.fdwrites))) {
+					// Double-check SelectForWrite() still returns true. If not, one of the callbacks must have
+					// modified some value since we checked SelectForWrite() earlier in this method.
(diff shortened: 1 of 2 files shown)


sodabrew:
@tmm1 Looks reasonable to me, any objections to merge?

ext/em.cpp
```diff
@@ -933,8 +933,12 @@ void EventMachine_t::_RunSelectOnce()
 					continue;
 				assert (sd != INVALID_SOCKET);
 
-				if (FD_ISSET (sd, &(SelectData.fdwrites)))
-					ed->Write();
+				if (FD_ISSET (sd, &(SelectData.fdwrites))) {
+					// Double-check SelectForWrite() still returns true. If not, one of the callbacks must have
+					// modified some value since we checked SelectForWrite() earlier in this method.
+					if (ed->SelectForWrite())
+						ed->Write();
```
tmm1: incorrect indentation

tmm1:
This feels like a hack, and will not work if you're using the epoll or kqueue reactors. That said, I don't have a better solution.

sodabrew:
@jim-slattery-rs Could I bug you to rebase this PR against master now that v1.0.5 is out?

jim-slattery-rs:
@sodabrew: sure thing, i just did the rebase

Issue #215, opened 2011-06-01
add wrapper for sendfile(2)


tmm1:
use sendfile(2) as a backend instead of `send_file_data` or `stream_file_data`.

PR #533, opened 2014-09-18
Integer overflow bug in fastfilereader and offset parameter to stream a file

mayanks:
Nothing too fancy. Just had a requirement to stream a file from a certain offset (in case the file was not streamed completely last time). So went ahead and added this feature in the eventmachine. 

Please accept this pull request if you feel it is a good feature to have.

### DIFF
--- a/ext/cmain.cpp
+++ b/ext/cmain.cpp
@@ -738,7 +738,7 @@ extern "C" int evma_set_rlimit_nofile (int nofiles)
 evma_send_file_data_to_connection
 *********************************/
 
-extern "C" int evma_send_file_data_to_connection (const unsigned long binding, const char *filename)
+extern "C" int evma_send_file_data_to_connection (const unsigned long binding, const char *filename, unsigned long offset)
 {
 	/* This is a sugaring over send_data_to_connection that reads a file into a
 	 * locally-allocated buffer, and sends the file data to the remote peer.
@@ -787,6 +787,9 @@ extern "C" int evma_send_file_data_to_connection (const unsigned long binding, c
 		close (Fd);
 		return -1;
 	}
+
+	// seek to the offset first
+	lseek(Fd, offset, SEEK_SET);
 
 	r = read (Fd, data, filesize);
 	if (r != filesize) {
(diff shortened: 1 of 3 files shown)


sodabrew:
Looks reasonable - please rebase the PR, and I think squash to a single commit, since it's basically the same set of changes in several places.

ext/cmain.cpp
```diff
@@ -785,6 +785,8 @@ extern "C" int evma_send_file_data_to_connection (const unsigned long binding, c
 		return -1;
 	}
 
+  // seek to the offset first
+  lseek(Fd, offset, SEEK_SET);
```
sodabrew: Mind the tabs, not spaces.

ext/fastfilereader/rubymain.cpp
```diff
@@ -68,8 +68,8 @@ static VALUE mapper_get_chunk (VALUE self, VALUE start, VALUE length)
 		rb_raise (rb_eException, "No Mapper Object");
 
 	// TODO, what if some moron sends us a negative start value?
-	unsigned _start = NUM2INT (start);
-	unsigned _length = NUM2INT (length);
+	long _start = NUM2LONG (start);
+	long _length = NUM2LONG (length);
```
sodabrew: These will be signed longs? Do you want to use NUM2ULONG?

sodabrew:
Hi - I noticed the added commits, but I don't think it came out as intended. You might have rebased locally, then done a 'git pull' and merged origin/master back into your local master before you pushed? The trick with a rebase is that you must force-push once you have the commits on your local branch organized as you wish, so that the origin matches local.

mayanks:
I should have commented that I messed this whole rebase business. I'll try and clean this up again. If I am not able to clean it, I'll submit a fresh pull request. I hope that should be fine.

mayanks:
Hi Aaron, after some hiccups and doing multiple reverts, I finally managed to clean the commit tree (and learned some new git tricks).

Do let me know if this looks good. Thanks for your patience.

sodabrew:
Yes, this looks much better! From reading this again, I agree with the NUM2ULONG changes. Actually, I wonder if those should be NUM2SIZET http://rxr.whitequark.org/mri/source/include/ruby/ruby.h#672

I am not comfortable with the offset parameter. The function `evma_send_file_data_to_connection` has a hardcoded 32KB buffer, so to send a large file you'd have to get the size, then iterate every 32KB until the end? That seems... less than ideal.

The best solution would be support for `sendfile()` per #215.

PR #534, opened 2014-09-18
RubyForge is no more

adelcambre:
Replace the dead documentation link in the README to rdoc.info which appears to be up to date canonical API docs.


jonbrenner:
The rdoc link on http://rubyeventmachine.com/ should also reference rdoc.info

sodabrew:
@jonbrenner Nice catch, thx. Pages rebuilt 9a1b7d7.

PR #535, opened 2014-09-19
instead of raising error, we can use `defined?`

kuldeepaggarwal:


sodabrew:
+1 this is better than using rescue for flow control.

Issue #538, opened 2014-09-23
t_invoke_popen should convert std::runtime_error exceptoins from evma_popen into Ruby exceptions


ericpromislow:
Otherwise the ruby process will stop with the unhelpful message "what: no fork"

Apps should be able to decide how to proceed when this happens.

sodabrew:
Resolved by 8711aa09022140d67de0ca1fdc6658734c280115

PR #539, opened 2014-09-23
Bug105194 fork crash

ericpromislow:
This addresses issue https://github.com/eventmachine/eventmachine/issues/538

I needed to patch package.rake as this is a branch off the 1.0.3 tag

You could simplify this by simply changing f to null in the exception handler,
but I wanted to return the errno value as well as the associated string.


sodabrew:
Could you rebase the PR now that eventmachine 1.0.4 has been released?

sodabrew:
Resolved by 8711aa09022140d67de0ca1fdc6658734c280115

PR #543, opened 2014-10-06
Handle iterator blocks returning Deferrable

singpolyma:
So that they can return Deferrable instead of calling iter.next
This also handles failure of one or more operations nicely

### DIFF
--- a/lib/em/iterator.rb
+++ b/lib/em/iterator.rb
@@ -41,6 +41,8 @@ module EventMachine
   #   end
   #
   class Iterator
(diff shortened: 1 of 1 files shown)


sodabrew:
Hi, could you explain the change and its use case here a little bit more?

singpolyma:
The idea is to have an iteration block that itself kicks off more asynchronous activities, and then move on when they're done instead of forcing the user to wire up the callback to the `iter.next` call manually.

PR #562, opened 2015-01-17
Fix file uploads in Windows.

mosoto:
In Windows the open function opens files in text mode by default which can cause problems with binary files.   See:
http://msdn.microsoft.com/en-us/library/z0kc8e3z.aspx
http://msdn.microsoft.com/en-us/library/ktss1a9b.aspx


sodabrew:
Looks reasonable to me.

Issue #496, opened 2014-02-20
Periodic Timer misfire or false fire


senthil-chinnaiyan:
We use eventmachine in an embedded linux platform (cross compiled). Some of the key features of network, timers, socket, file/directory watch and differ are used heavily. We have few issues related to the system time change and they are:

1. When the System Time is set to back date, the periodic timer misfires.
2. Ehen the System time is advanced, then periodic timer false fires.
3. Sometimes the CPU is touching very high percentage and after some time the ruby process is put into zombie state. (There is a fix for this: https://github.com/eventmachine/eventmachine/issues/428), but this doesn’t solve the above 2 issues.

To re-produce:
1. Set the time of the linux system using date command: date +%T -s "16:36:30"
2. Run this code:

require "eventmachine"

EM.run {
    EventMachine::PeriodicTimer.new(2) do
        puts "Timer Fired: #{Time.now}"
    end
}
3. Generates output like this:
Timer Fired: 2014-10-21 16:36:35 -0500
Timer Fired: 2014-10-21 16:36:37 -0500
Timer Fired: 2014-10-21 16:36:39 -0500
Timer Fired: 2014-10-21 16:36:41 -0500
4. Change the time to backward (10 seconds) in another console: date +%T -s "16:36:20"
5. No output for 10 seconds and the timer fires after the system time is greater than next fire_at time.
 
After investigation, we have found that the _GetTimers() function skips the fire because the MyCurrentLoopTime  is lesser than the i->first. Reason being the MyCurrentLoopTime is associated with the system clock.

Solution (Tested in Linux):
To resolve this, the gettimeofday should be avoided and clock_gettime(CLOCK_MONOTONIC, &ts_spec); can be used instead. This will ensure that the MyCurrentLoopTime won’t be affected if the system time is changed.

Tested on Linux and works as expected.

uint64_t EventMachine_t::GetRealTime()
{
	uint64_t current_time;

	#if defined(OS_UNIX)
                /*struct timeval tv;*/
                /*gettimeofday (&tv, NULL);*/
                /*current_time = (((uint64_t)(tv.tv_sec)) * 1000000LL) + ((uint64_t)(tv.tv_usec));*/

                struct timespec tv;
                clock_gettime(CLOCK_MONOTONIC, &tv);
                current_time = (((uint64_t)(tv.tv_sec)) * 1000000LL) + ((uint64_t)((tv.tv_nsec)/1000));

	#elif defined(OS_WIN32)
[...]

PR #551, opened 2014-12-01
Update build deps

sodabrew:
Working through Travis failures seen in 3349891

PR #563, opened 2015-01-18
Use clock_gettime and mach_absolute_time to make GetRealTime monotonic

sodabrew:
Based on clock_gettime from #496 with support for Mac OS X based on https://developer.apple.com/library/mac/qa/qa1398/_index.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,6 @@ matrix:
   allow_failures:
     - rvm: rbx
     - rvm: jruby
+  include:
+    - rvm: 2.0.0
+      os: osx
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -85,6 +85,11 @@ EventMachine_t::EventMachine_t (EMCallback event_callback):
 	Quantum.tv_sec = 0;
 	Quantum.tv_usec = 90000;
 
+	/* Initialize monotonic timekeeping on OS X before the first call to GetRealTime */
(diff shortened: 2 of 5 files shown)


rud:
Travis did not like the commit, is it common for the tests to hang?

sodabrew:
Yes, I am working in #551 to resolve test problems.

sodabrew:
Just found _the best_ article I've read on this topic: http://nadeausoftware.com/articles/2012/04/c_c_tip_how_measure_elapsed_real_time_benchmarking
I'm going to update this PR to use the same `#if defined(CLOCK_FOO)` pattern rather than `#if defined(OS_FOO)` pattern.

sodabrew:
Turns out that RubyInstaller targets Windows XP as its baseline Win32 API, which is why I had to do the gymnastics for windows.h to give me GetTickCount64(), which means this wouldn't compile on XP. So I'll drop that for now with a note to use the newer function in the future. Another rebase is pending.

sodabrew:
I have tested this on a few platforms now:
Linux (CLOCK_MONOTONIC_RAW)
FreeBSD 10 (CLOCK_MONOTONIC)
Mac OS X (mach_absolute_time)
Windows 8.1 (GetTickCount)

I'm in the middle of downloading a new Solaris 11 VM image.

sodabrew:
Merged #551, rebased this, now we will get Travis unit tests.

PR #568, opened 2015-01-27
Rubinius popen and test fixes

sodabrew:
The Rubinius work from #551 is rebased here.

### DIFF
--- a/tests/em_test_helper.rb
+++ b/tests/em_test_helper.rb
@@ -45,6 +45,10 @@ class Test::Unit::TestCase
     def jruby?
       defined? JRUBY_VERSION
     end
+
+    def rbx?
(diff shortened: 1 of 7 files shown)


sodabrew:
I've decided that I'm happy with this with respect to rubinius/rubinius#3268.

tmm1:
Diff contains a swp file

sodabrew:
Derp. Removed swp file.

tests/em_test_helper.rb
```diff
@@ -45,6 +45,10 @@ def windows?
     def jruby?
       defined? JRUBY_VERSION
     end
+
+    def rbx?
```
brixen: You may be interested in https://github.com/brixen/redcard

Issue #123, opened 2010-11-08
Exception when scheduling a block after reactor is shut down (Java)


dougbradbury:
I came across this exception when I tried to schedule a block and my reactor had already been shut down because of another (unrelated) error.  

```
gems/eventmachine-0.12.11-java/lib/jeventmachine.rb:142:in `signal_loopbreak'
gems/eventmachine-0.12.11-java/lib/eventmachine.rb:1042:in `next_tick'
gems/eventmachine-0.12.11-java/lib/eventmachine.rb:257:in `schedule'
pcres/management_console/evented_connection.rb:12:in `send_message'
pcres/management_console/evented_connection.rb:83:in `send_message'
pcres/server/mc_session.rb:157:in `send_message'
pcres/server/chats/chat.rb:109:in `send_message'
pcres/server/chats/next_reservation_chat.rb:17:in `start'
pcres/server/chat_runner.rb:36:in `run'
pcres/server/mc_session.rb:69:in `initialize'
pcres/server/mc_session.rb:69:in `new'
pcres/server/mc_session.rb:69:in `initialize'
pcres/server/session_builder.rb:53:in `up_to_date_actions'
pcres/server/session_builder.rb:20:in `hello'
gems/dkbrpc-0.1.11/lib/dkbrpc/incoming_connection.rb:15:in `receive_message'
gems/dkbrpc-0.1.11/lib/dkbrpc/fast_message_protocol.rb:48:in `receive_data'
gems/eventmachine-0.12.11-java/lib/eventmachine.rb:1397:in `event_callback'
gems/eventmachine-0.12.11-java/lib/jeventmachine.rb:83:in `eventCallback'
gems/eventmachine-0.12.11-java/lib/jeventmachine.rb:102:in `run_machine'
gems/eventmachine-0.12.11-java/lib/eventmachine.rb:195:in `run'
pcres/server/main.rb:69:in `start_eventmachine'
pcres/server/main.rb:68:in `initialize'
pcres/server/main.rb:68:in `new'
pcres/server/main.rb:68:in `start_eventmachine'
pcres/server/main.rb:27:in `run'<script>:1
```

PR #572, opened 2015-02-02
Abstract out an EmReactorInterface...

sodabrew:
Create a NullEmReactor and NullEventableChannel to use when the reactor should not be running.

This is a rebase of PR #124.

I have no idea if this is still needed.

### DIFF
--- a/java/src/com/rubyeventmachine/EmReactor.java
+++ b/java/src/com/rubyeventmachine/EmReactor.java
@@ -36,7 +36,9 @@ import java.net.*;
 import java.util.concurrent.atomic.*;
 import java.security.*;
 
-public class EmReactor {
+public class EmReactor implements EmReactorInterface
+{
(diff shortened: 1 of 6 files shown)


sodabrew:
Will need to come back to this:
```
java/src/com/rubyeventmachine/EmReactor.java:39: error: EmReactor is not abstract and does not override abstract method sendDatagram(long,String,int,String,int) in EmReactorInterface
public class EmReactor implements EmReactorInterface
       ^
java/src/com/rubyeventmachine/NullEventableChannel.java:7: error: NullEventableChannel is not abstract and does not override abstract method getSockName() in EventableChannel
public class NullEventableChannel implements EventableChannel
       ^
```

sodabrew:
The problem from #123 is still reproducible, this does fix it. Merging for 1.3 series.

Issue #482, opened 2013-11-28
CPU and memory usage turns high when EM.run in rescue block


halida:
here is the code:

```ruby
require 'eventmachine'

def run
  EventMachine.run do
    EM.next_tick{ puts "12" }
  end
end

def main
  begin
    raise
  rescue Exception
    run
  end
end

main
```

After run this code, memory usage gradually increase, and CPU usage is high.
I'm using gdb to check this out, found lots of Proc object created, and this Proc object is this one in run_deferred_callbacks :
https://github.com/eventmachine/eventmachine/blob/cfb1f71a35b1a10e5821bca9841fee3080ec1685/lib/eventmachine.rb#L975

My enviroment:
OSX 10.9,
rvm + ruby 2.0.0-p353
eventmachine (1.0.3)

sodabrew:
This sounds a lot like #452. I wasn't able to reproduce so I abandoned #573, but I will take another look at it.

sodabrew:
Resolved by #452

PR #573, opened 2015-02-02
Add test for EM.schedule from within a rescue block

sodabrew:
Test for #452


sodabrew:
Yup, sure does hang on 1.8.7.

sodabrew:
Actually I can't repro this anymore.

sodabrew:
Unfortunately I wasn't able to put together a unit test for this problem, but there are repro steps in #482 and #452.

Issue #575, opened 2015-02-03
ThreadedResource dispatch test fails in Ruby 2.1, 2.2, and Rubinius


sodabrew:
This test fails frequently on Ruby 2.1, 2.2, and Rubinius. It looks like a race condition.

```
TestThreadedResource: 
  test_dispatch_completion:             E
===============================================================================
Error: test_dispatch_completion(TestThreadedResource)
  RuntimeError: Resource dispatch timed out
/home/travis/build/eventmachine/eventmachine/tests/test_threaded_resource.rb:20:in `block (2 levels) in test_dispatch_completion'
/home/travis/build/eventmachine/eventmachine/lib/eventmachine.rb:187:in `call'
/home/travis/build/eventmachine/eventmachine/lib/eventmachine.rb:187:in `run_machine'
/home/travis/build/eventmachine/eventmachine/lib/eventmachine.rb:187:in `run'
/home/travis/build/eventmachine/eventmachine/tests/test_threaded_resource.rb:17:in `test_dispatch_completion'
     14:   end
     15: 
     16:   def test_dispatch_completion
  => 17:     EM.run do
     18:       EM.add_timer(3) do
     19:         EM.stop
     20:         fail 'Resource dispatch timed out'
===============================================================================
: (3.005189)
```

I added the timeout after 3 seconds to prevent the unit test suite from hanging indefinitely, so what this means is that the dispatch is never dispatching. The test code is in `tests/test_threaded_resource.rb`:

``` ruby
  def test_dispatch_completion
    EM.run do
      EM.add_timer(3) do
        EM.stop
        fail 'Resource dispatch timed out'
      end
      completion = resource.dispatch do |o|
        o[:foo] = :bar
        :foo
      end
      completion.callback do |result|
        assert_equal :foo, result
        EM.stop
      end
      completion.errback do |error|
        EM.stop
        fail "Unexpected error: #{error.message}"
      end
    end
    assert_equal :bar, object[:foo]
  end
```

voxik:
I might be just lucky, but from approx 20 test runs of 1.0.8 on Ruby 2.3, I have not  met this issue yet.

Will try to enable this test in Fedora's build of EM. You can observe the CI builds here [1], particularly the build.log files.

[1] https://apps.fedoraproject.org/koschei/package/rubygem-eventmachine

sodabrew:
I'm no longer seeing this.

PR #576, opened 2015-02-03
Add puts messages to test_threaded_resource

sodabrew:
Attempting to debug #575.


sodabrew:
And this makes it work reliably. Five bucks says `puts` creates implicit locking by way of `write(STDOUT, ...)`.

sodabrew:
I'm no longer seeing this.

PR #583, opened 2015-02-06
Honour requests to close NOW

zimbatm:
...even if we have previously asked to close later. Many places in the
codebase use ScheduleClose(false) when they have found out that the
socket is no longer operational, but they are being ignored if
ScheduleClose(true) has been used in the past. As the socket is often
not working at this point, we will never get through the send queue and
actually do the close.

### DIFF
--- a/ext/ed.cpp
+++ b/ext/ed.cpp
@@ -217,8 +217,14 @@ EventableDescriptor::ScheduleClose
 
 void EventableDescriptor::ScheduleClose (bool after_writing)
 {
-	if (IsCloseScheduled())
+	if (IsCloseScheduled()) {
+		if (!after_writing) {
+			// If closing has become more urgent, then upgrade the scheduled
+			// after_writing close to one NOW.
+			bCloseNow = true;
+		}
 		return;
+	}
 	MyEventMachine->NumCloseScheduled++;
 	// KEEP THIS SYNCHRONIZED WITH ::IsCloseScheduled.
 	if (after_writing)


sodabrew:
A unit test would be awesome - is there a specific failure mode that we can model in a test?

zimbatm:
I had a go at it yesterday but eventmachine is segfaulting too much on master.
https://gist.github.com/zimbatm/b7998fd4f82a039d2c9c

tmm1:
Can you include the Crash Report in that gist?

zimbatm:
Done

zimbatm:
Ruby is installed using `ruby-install`. I've tested on ruby 2.2 and had similar issues.

tmm1:
Strange, I've never seen that segfault before. Can you try this patch?

``` diff
diff --git a/ext/em.cpp b/ext/em.cpp
index 3bcaa4a..1b4fc34 100644
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -132,6 +132,9 @@ EventMachine_t::~EventMachine_t
 
 EventMachine_t::~EventMachine_t()
 {
+	// Clear heartbeats
+	Heartbeats.clear();
+
 	// Run down descriptors
 	size_t i;
 	for (i = 0; i < NewDescriptors.size(); i++)

```

zimbatm:
Thanks for looking into this. Here's some more segfault after applying the patch: https://gist.github.com/zimbatm/feb0a39e6863fbdeef55

Issue #585, opened 2015-02-09
Eventmachine >= 1.0.5 with Thin on Windows with Ruby 2.x leaking memory


neongrau:
Observed with 1.0.6: as soon as the ruby process is started it'll leak 16KB chunks of memory pretty much every second. This happens without serving any Requests (load balancer in front was disabled).

After downgrading to 1.0.4 this does not happen anymore. Quick test with Eventmachine 1.0.5 also had the memory leak.

virtualfunction:
Interesting, I noticed a similar issue, but tried rolling back and found the issue still persisted. I'm using EM with the Sinatra SSE gem which happens to add timers for each connection to send keep alive data. This seemed to be the source of the leak. I ditched the SSE gem and used a static app level timer instead of a connection level one to remove the mem leaks.

tmm1:
@neongrau Are you using SSL? Or any of the other features that were changed in the 1.0.5 release: https://github.com/eventmachine/eventmachine/blob/master/CHANGELOG.md#105-february-2-2015 ?

neongrau:
Not using SSL on that side. When i use SSL it happens on an Apache Load Balancer.
It's quite easy to reproduce. Create a fresh Rails app "rails new testapp" add "gem 'thin'" to Gemfile and do "bundle install".
With "thin start" you will immediatly see the memory ticking up.

I'm not sure if that usage scenario corresponds to any of the changes in 1.0.5

tmm1:
@neongrau I think I found the leak. Can you confirm which version of ruby, platform (linux/mac/windows) and whether you are using kqueue or epoll?

neongrau:
Most of my testing was with Ruby 2.0.0p598 x64-mingw32. But also with 2.1.5 which made no difference except even higher memory usage. Can't really answer the kqueue/epoll question. Just using Thin 1.6.0 - 1.6.3 which is the part that relies on Eventmachine.

tmm1:
@neongrau Can you try v1.0.7, it should fix this issue.

neongrau:
Yes, 1.0.7 seems to have fixed the leak. Now starting a longer running test as i had another issue with the Thin server stop responding to requests after a couple of hours.

le0pard:
Confirmed. 

Memory leak was in v1.0.6 and it disappear after update to v1.0.7. 

http://monosnap.com/image/z6UbbSozx7xwnSv1cqU4GWVeMtwsci

PR #586, opened 2015-02-10
Fix fdset memory leak

tmm1:
fixes leak reported in https://github.com/eventmachine/eventmachine/issues/585#issuecomment-73646276

originally introduced in https://github.com/eventmachine/eventmachine/pull/502

/cc @piki @sodabrew

### DIFF
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -882,7 +882,7 @@ SelectData_t::~SelectData_t()
 _SelectDataSelect
 *****************/
 
-#if !defined(HAVE_RB_THREAD_FD_SELECT) && (defined(HAVE_TBR) || defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL))
+#if defined(HAVE_TBR) || defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL)
 static VALUE _SelectDataSelect (void *v)
 {
 	SelectData_t *sd = (SelectData_t*)v;
@@ -897,10 +897,7 @@ SelectData_t::_Select
 
 int SelectData_t::_Select()
 {
-	#if defined(HAVE_RB_THREAD_FD_SELECT)
-	// added in ruby 1.9.2
-	return rb_thread_fd_select (maxsocket+1, &fdreads, &fdwrites, &fderrors, &tv);
-	#elif defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL)
+	#if defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL)
 	// added in ruby 1.9.3
 	rb_thread_call_without_gvl ((void *(*)(void *))_SelectDataSelect, (void*)this, RUBY_UBF_IO, 0);
 	return nSockets;
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -879,7 +879,7 @@ SelectData_t::~SelectData_t()
 _SelectDataSelect
 *****************/
 
-#if defined(HAVE_TBR) || defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL)
+#if !defined(HAVE_RB_THREAD_FD_SELECT) && (defined(HAVE_TBR) || defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL))
 static VALUE _SelectDataSelect (void *v)
 {
 	SelectData_t *sd = (SelectData_t*)v;
@@ -894,10 +894,15 @@ SelectData_t::_Select
 
 int SelectData_t::_Select()
 {
-	#if defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL)
+	#if defined(HAVE_RB_THREAD_FD_SELECT)
+	// added in ruby 1.9.2
+	return rb_thread_fd_select (maxsocket+1, &fdreads, &fdwrites, &fderrors, &tv);
+	#elif defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL)
+	// added in ruby 1.9.3
 	rb_thread_call_without_gvl ((void *(*)(void *))_SelectDataSelect, (void*)this, RUBY_UBF_IO, 0);
 	return nSockets;
 	#elif defined(HAVE_TBR)
+	// added in ruby 1.9.1, deprecated in ruby 2.0.0
 	rb_thread_blocking_region (_SelectDataSelect, (void*)this, RUBY_UBF_IO, 0);
 	return nSockets;
 	#else
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -911,7 +911,12 @@ int SelectData_t::_Select()
 }
 #endif
 
-
+void SelectData_t::_Clear()
+{
+	rb_fd_zero (&fdreads);
+	rb_fd_zero (&fdwrites);
+	rb_fd_zero (&fderrors);
+}
 
 /******************************
 EventMachine_t::_RunSelectOnce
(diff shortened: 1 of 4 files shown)


ext/em.cpp
```diff
@@ -592,12 +592,12 @@ void EventMachine_t::_RunEpollOnce()
...
-	rb_fdset_t fdreads;
+	fd_set fdreads;
 
-	rb_fd_init(&fdreads);
-	rb_fd_set(epfd, &fdreads);
+	FD_ZERO(&fdreads);
+	FD_SET(epfd, &fdreads);
 
-	if ((ret = rb_thread_fd_select(epfd + 1, &fdreads, NULL, NULL, &tv)) < 1) {
+	if ((ret = rb_thread_select(epfd + 1, &fdreads, NULL, NULL, &tv)) < 1) {
```
tmm1: This epoll (and the below kqueue) path is only for legacy rubies that only shipped with `rb_thread_select`. Reactors are generally started very early during a ruby process, so it is highly unlikely that an epoll/kqueue fd will ever exceed FD_SETSIZE. It's safe to use a regular stack allocate fd_set here.

ext/em.cpp
```diff
@@ -877,7 +883,7 @@ _SelectDataSelect
 static VALUE _SelectDataSelect (void *v)
 {
 	SelectData_t *sd = (SelectData_t*)v;
-	sd->nSockets = rb_fd_select (sd->maxsocket+1, &(sd->fdreads), &(sd->fdwrites), &(sd->fderrors), &(sd->tv));
+	sd->nSockets = select (sd->maxsocket+1, &(sd->fdreads), &(sd->fdwrites), &(sd->fderrors), &(sd->tv));
```
tmm1: This has to use a raw `select` because it is invoked inside a "thread blocking region". This was a legacy api in ruby 1.9.x which allowed callbacks to be invoked outside ruby's GVL, and thus is not allowed to call back into any ruby apis.

ext/em.cpp
```diff
@@ -1045,6 +1051,7 @@ void EventMachine_t::_CleanBadDescriptors()
 		rb_fd_set(sd, &fds);
 
 		int ret = rb_fd_select(sd + 1, &fds, NULL, NULL, &tv);
+		rb_fd_term(&fds);
```
tmm1: This codepath is only exercised when a BADF is encountered via the select() reactor. It is quite rare, so I left the heap allocation in place and am explicitly cleaning up here.

ext/em.cpp
tmm1: This is failing to compile since we're no longer using raw `fd_set`s. Switching to `select` here also reintroduces the original issue where the select reactor breaks on large fds.

Unfortunately it doesn't look like rb_fd_select is safe to call from outside the gvl, because it can invoke ruby_xmalloc which can kick off a ruby GC.

The correct solution here is probably to prefer `rb_fd_select` over the raw `rb_thread_blocking_region` and `rb_thread_call_without_gvl` apis when available, to let the VM do all the blocking GVL stuff for us.

ext/em.h
```diff
@@ -22,7 +22,12 @@ See the file COPYING for complete licensing information.
 
 #ifdef BUILD_FOR_RUBY
   #include <ruby.h>
-  #define EmSelect rb_thread_fd_select
+  #ifdef HAVE_RB_THREAD_FD_SELECT
+    #define EmSelect rb_thread_fd_select
+  #else
+    // ruby 1.9.1 and below
+    #define EmSelect rb_thread_select
```
tmm1: 1.9.1 is a special case because it defines rb_fdset_t and rb_thread_select, but not rb_fd_init or rb_thread_fd_select

ext/em.cpp
```diff
@@ -867,17 +870,23 @@ SelectData_t::SelectData_t()
 	rb_fd_init (&fderrors);
 }
 
+SelectData_t::~SelectData_t()
+{
+	rb_fd_term (&fdreads);
+	rb_fd_term (&fdwrites);
+	rb_fd_term (&fderrors);
```
tmm1: These `rb_fd_term` will cleanup heap allocated memory associated with the rb_fdset_t, but I wanted to avoid allocation/deallocation on every reactor tick so I made the entire SelectData_t get allocated once when the reactor is created.

tmm1:
cc @neongrau this should fix the memory leak you are seeing

tmm1:
This leak was present on any ruby >1.9.1 when using the select reactor, and only on ruby <2.0.0 when using the epoll or kqueue reactor. Our production epoll+ruby 2.1 instances were not affected which is why I didn't notice the memory leak.

ext/em.cpp
piki: Thanks for all the comments inline.  This one is especially helpful, because it explains something that looks at first glance like a reversion.

:+1: to all of it.

SamSaffron:
Looks like we also saw this leak pretty bad at Discourse.

tmm1:
We should probably yank the bad releases from rubygems /cc @sodabrew

SamSaffron:
@tmm1 on one hand some people say, never ever yank gems, on the other hand I think it is called for here because it is very hard to debug, breaking deployments is actually desirable here. 

cc @indirect

indirect:
The current tradeoffs mean I only recommend it for security issues. Yanking will break new deploys or checkouts for every project that depends on the yanked gem.

SamSaffron:
@indirect a memory leak that will take out your server at 3AM in the morning while you are sleeping is arguably just as important as a security flaw, perhaps even more. 

Its a tough call to make, but I think "security only" should be a policy that should have a bit of movement for cases such as this, especially since it is so hard to diagnose.

indirect:
@SamSaffron well, that's why it's my recommendation rather than my demand. :)

More seriously, it's a tradeoff. Some people are apparently using this version without any problems, and those people will have their ability to develop or deploy arbitrarily broken by a yank. Other people have a potential problem of a memory leak taking down their app servers. I have no idea which group of people is bigger or more important, but my general advice is (and has been) to not yank and ask people to upgrade instead, in order to reduce breakage.

SamSaffron:
@indirect I totally get where you are getting at. 

The thing that is really hurting our ecosystem is that there is no way of communicating this kind of stuff outside of yanking. 

Sure, Aman can tweet about it, but at best it will reach a fraction of the users. The homepage here could be updated but really nobody is going to the readme for news on a regular basis, when you are lugging around 200 dependencies its just not practical to subscribe to every stream of information. 

What we really need is a way to be able to simply flag a gem as "problematic", then next "bundle install" can output `your gem blaba has a serious memory leak, please upgrade to version 2.3 urgently` , it would be super duper awesome and cover this issue just fine. It is also way better than yanking cause it would teach you why?

indirect:
Notifications about gems seem good. Someone should build that. :)

SamSaffron:
@indirect Google Summer Of Code !!! :+1:  does bundler have a project there, only 4 days left

sodabrew:
At the moment, there have been [45,324 downloads](https://rubygems.org/gems/eventmachine/versions/1.0.6
) of 1.0.6 and [27,480 downloads](https://rubygems.org/gems/eventmachine/versions/1.0.5
) of 1.0.5. If those numbers are still moving significantly in a few days, I would consider a yank. 

+1 for adding notices to the Rubygems API, that'd be really neat. :seedling:

SamSaffron:
update on @sodabrew's number

72183 downloads for 1.0.5 
57421 downloads for 1.0.6 

I recommend yanking @tmm1  this bleeding should stop. It is better to break builds than have them deal with debugging this.

sodabrew:
Compared to 350,000 downloads for 1.0.7 after ~45 days. I don't feel too strongly either way. @SamSaffron it sounds like you feel strongly for the yank?

SamSaffron:
I think its appropriate for this particular case. your call

tmm1:
Yea let's yank them.

PR #587, opened 2015-02-10
Signal loopbreaker during halt

### LINKED ISSUE #522: When using epoll, `stop_event_loop` doesn't take effect until the next timer fires
When eventmachine is using `EM.epoll` and has at least one timer set up, a call to `stop_event_loop` will have no effect until the next timer fires.

For example, this program takes up to five seconds to exit when it gets a SIGTERM.  If there aren't any timers, the program exits after about 90ms, the value of `Quantum.tv_usec`.  If the `EM.epoll` line is commented out, the program exits immediately, whether or not any timers are scheduled.

```
EM.epoll
EM.run {
  EM.add_periodic_timer(5) do
    puts "timer"
  end
  EM.add_shutdown_hook do
    puts "shutting down: #{Time.now.to_f}"
  end
  Signal.trap("TERM") do
    puts "SIGTERM: #{Time.now.to_f}"
    EM::stop_event_loop
  end
}
```

Does the epoll version of `stop` need to signal the loop breaker?

tmm1:
fixes #522 

confirmed `kill -TERM` works against a kqueue reactor loop after this patch using the repro in https://github.com/eventmachine/eventmachine/issues/522#issue-37535399

```
tmm1@tmm1-air ~/code/eventmachine (loopbreak-at-halt*) $ ruby -Ilib -r eventmachine test.rb
30227
timer
SIGTERM: 1423556870.274626
shutting down: 1423556870.274733
```

/cc @piki

### DIFF
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -208,6 +208,12 @@ void EventMachine_t::ScheduleHalt()
    * The answer is to call evma_stop_machine, which calls here, from a SIGINT handler.
    */
 	bTerminateSignalReceived = true;
+
+	/* Signal the loopbreaker so we break out of long-running select/epoll/kqueue and
+	 * notice the halt boolean is set. Signalling the loopbreaker also uses a single
+	 * signal-safe syscall.
+	 */
+	SignalLoopBreaker();
 }
 
 


piki:
Glad the simple fix worked.  Thanks!

PR #591, opened 2015-04-16
Open datagram sockets with CLOEXEC.

robx:
So as not to share them with children, and in particular not fail
to reopen on restart if a child is still around.

This is meant to fix a sensu-client issue (https://github.com/sensu/sensu/issues/862) where the process fails to come back up due to a child that keeps the socket.

### DIFF
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -1679,6 +1679,16 @@ const unsigned long EventMachine_t::OpenDatagramSocket (const char *address, int
 			goto fail;
 	}
 
+	// Set CLOEXEC. Only makes sense on Unix.
+	{
+		#ifdef OS_UNIX
+		int cloexec = fcntl (sd, F_GETFD, 0);
+		assert (cloexec >= 0);
+		cloexec |= FD_CLOEXEC;
+		fcntl (sd, F_SETFD, cloexec);
+		#endif
+	}
+
 	if (bind (sd, (struct sockaddr*)&sin, sizeof(sin)) != 0)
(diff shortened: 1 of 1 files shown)


sodabrew:
Thanks for this. Actually I would like to refactor this a bit including setting the non-blocking status on the socket, to look more like this from mysql2 (adapted to `socket` rather than `open` but the same pattern. The majority of systems can accept SOCK_NONBLOCK and SOCK_CLOEXEC in the `socket` call.

``` C
#ifdef SOCK_CLOEXEC
  /* Atomically set CLOEXEC on the new FD in case another thread forks */
  int sockfd = open("/dev/null", O_RDWR | O_CLOEXEC);
  if (sockfd < 0) {
    /* Maybe SOCK_CLOEXEC is defined but not available on this kernel */
    int sockfd = open("/dev/null", O_RDWR);
    fcntl(sockfd, F_SETFD, FD_CLOEXEC);
  }
#else
  /* Well we don't have SOCK_CLOEXEC, so just set FD_CLOEXEC quickly */
  int sockfd = open("/dev/null", O_RDWR);
  fcntl(sockfd, F_SETFD, FD_CLOEXEC);
#endif
```

https://github.com/brianmario/mysql2/blob/master/ext/mysql2/client.c#L173-185

sodabrew:
... on further reading, SO_CLOEXEC is already being set in CreateTcpServer, and actually you're using the same block of code here.

sodabrew:
Merged, and also superseded by #488.

PR #600, opened 2015-06-10
SmtpClient: Properly escape leading periods in DATA

davidbalbert:
The DATA command of the SMTP protocol is terminated by a `\r\n.\r\n`. In order to allow for lines starting with a period or containing only a period, the protocol defines an escape mechanism. From [Section 4.5.2 of RFC5321](https://tools.ietf.org/html/rfc5321#section-4.5.2):

> Before sending a line of mail text, the SMTP client checks the first character of the line.  If it is a period, one additional period is inserted at the beginning of the line.

`EM::Protocols::SmtpClient` currently doesn't do this, and messages that contain the string `\r\n.\r\n` somewhere in the body will be truncated by SMTP servers. This is something we have encountered in the wild.

This PR does a few things:

1. Add `:message` as an additional option to SMTP client. This accepts a string that is a valid RFC2822 Internet Message. When using this option, `SmtpClient` will escape leading periods and add a trailing `\r\n.\r\n`.

2. Escape leading periods in `@args[:headers]` and `@args[:body]`. `@args[:content]` is not escaped because it expects data already in the proper format.

3. Update documentation:
  - Add documentation for `:message`
  - Remove documentation for `:content` being specified as an array, which isn't currently supported.
  - Specify that `:content` must included escaped leading periods in addition to a trailing `\r\n.\r\n`.
  - Replace MailFactory example with Mail example.
  - Remove outdated todo.

### DIFF
--- a/lib/em/protocols/smtpclient.rb
+++ b/lib/em/protocols/smtpclient.rb
@@ -48,20 +48,20 @@ module EventMachine
     #     puts 'Email failed!'
     #   }
     #
-    # Sending generated emails (using mailfactory)
+    # Sending generated emails (using Mail)
     #
-    #   mail = MailFactory.new
-    #   mail.to = 'someone@site.co'
-    #   mail.from = 'me@site.com'
-    #   mail.subject = 'hi!'
-    #   mail.text = 'hello world'
-    #   mail.html = '<h1>hello world</h1>'
+    #   mail = Mail.new do
+    #     from    'alice@example.com'
+    #     to      'bob@example.com'
+    #     subject 'This is a test email'
+    #     body    'Hello, world!'
+    #   end
     #
     #   email = EM::P::SmtpClient.send(
-    #     :domain=>'site.com',
-    #     :from=>mail.from,
+    #     :domain=>'example.com',
+    #     :from=>mail.from.first,
     #     :to=>mail.to,
-    #     :content=>"#{mail.to_s}\r\n.\r\n"
+    #     :message=>mail.to_s
     #   )
     #
     class SmtpClient < Connection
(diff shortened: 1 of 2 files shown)


davidbalbert:
cc @zachallaun who worked on this as well.

davidbalbert:
We've fixed the tests on 1.8.7. Please let me know if you'd like us to squash the commits.

icco:
TIL. :+1:

sodabrew:
Thank you for submitting this! Please do squash.

davidbalbert:
@sodabrew squashed! Thanks for responding so quickly.

davidbalbert:
@sodabrew any updates on this? Let me know if there's anything else you need from  me.

sodabrew:
I read it over more carefully - the prior help text suggests that `:body` needs to be fully escaped, and looks forward to a `:content` parameter that will do the escaping for you. In your diff, these appear to be swapped semantically, with `:body` being the "loglcal" value and `:content` the "raw" value.

zachallaun:
@sodabrew (Stepping in for @davidbalbert here – we worked on this PR together.)

I believe you're referring to the following diff (truncated):

```diff
-      # :body => Optional string, defaults blank.
+      # :body => Optional String or Array of Strings, defaults blank.
       #   This will be passed as the body of the email message.
       #   TODO, this needs to be significantly beefed up. As currently written, this requires the caller
       #   to properly format the input into CRLF-delimited lines of 7-bit characters in the standard
       #   SMTP transmission format. We need to be able to automatically convert binary data, and add
-      #   correct line-breaks to text data. I think the :body parameter should remain as it is, and we
-      #   should add a :content parameter that contains autoconversions and/or conversion parameters.
-      #   Then we can check if either :body or :content is present and do the right thing.
...
+      #   correct line-breaks to text data.
```

We believed this comment to be outdated. `:content` already existed as a parameter, and it was being sent raw without any escaping. The handling of the `:content` case is unchanged in our patch:

```diff
-        if @args[:content]
+        elsif @args[:content]
           send_data @args[:content].to_s
```

The majority of our comment changes not regarding the new `:message` parameter are clarifications and corrections, e.g. the outdated comment above, and suggesting that `:content` could be an array of strings (as seen in the diff above, it cannot be).

The only change we made to the previous parameters was that leading dots in `:body` will now be escaped. This is not backwards-incompatible, however, because our escaping method is a no-op if the leading dots are already escaped.

Please let me know if I misunderstood you or am missing something, and thanks for going over this with us.

sodabrew:
Ok, I get it now. Thanks for clarifying and pointing me in the right direction at the diff.

PR #617, opened 2015-07-30
Crank the machine from Ruby space when running Rubinius

### LINKED ISSUE #201: memory leak with EventMachine::Connection objects on Rubinius
This seems to be similar to #157, but I can't get the objects to collect ever.

Using about as simple a server as I can:

```ruby
require 'rubygems'
require 'eventmachine'

module Echo
  def receive_data(data)
    send_data data
  end
end

EventMachine.run do
  EventMachine.add_periodic_timer(5) do
    print "EventMachine::Connection objects: "
    puts ObjectSpace.each_object(EventMachine::Connection) { }
  end

  EventMachine.start_server '127.0.0.1', 4015, Echo
end
```

If you run this on MRI 1.8.7 or MRI 1.9.2 and connect to it several hundred times (and close those connections) the objects behave strangely. Some are collected, but the vast majority hang around forever. If you force a `GC.start` in the timer, all but two are collected.

If you run this on Rubinius they are never collected, even with `GC.start`. I have also filed a bug report with Rubinus folks, but since they're never collected on MRI without a forced GC run I thought I'd file it here too.

sodabrew:
This is a bigger refactor than #202 but I think it is cleaner overall.

Resolves #201, #202.

### DIFF
--- a/lib/eventmachine.rb
+++ b/lib/eventmachine.rb
@@ -184,7 +184,15 @@ module EventMachine
           add_timer(0) { signal_loopbreak }
         end
         @reactor_thread = Thread.current
-        run_machine
+
+        # Rubinius needs to come back into "Ruby space" for GC to work,
+        # so we'll crank the machine here.
+        if defined?(RUBY_ENGINE) && RUBY_ENGINE == "rbx"
+          while run_machine_once; end
+        else
+          run_machine
+        end
+
       ensure
         until @tails.empty?
           @tails.pop.call
--- a/ext/cmain.cpp
+++ b/ext/cmain.cpp
@@ -78,6 +78,17 @@ extern "C" void evma_release_library()
 }
 
 
+/*********************
+evma_run_machine_once
+*********************/
+
+extern "C" bool evma_run_machine_once()
+{
+	ensure_eventmachine("evma_run_machine_once");
+	return EventMachine->RunOnce();
+}
+
+
 /****************
 evma_run_machine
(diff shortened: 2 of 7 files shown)


sodabrew:
Looks like moving the epoll / kqueue loopbreakers from Run into the constructor isn't working, hmm.

sodabrew:
Figured out that if I set up the loopbreakers from the constructor, then I also need to know the poller-type at construction time.

sodabrew:
Renamed Crank back to RunOnce, which makes the diff more concise. There was already a private _RunOnce, so it's just public now as RunOnce, along with refactoring the constructor.
(Crank last seen in d84c841 for amusement).

lib/eventmachine.rb
```diff
@@ -184,7 +184,15 @@ def self.run blk=nil, tail=nil, &block
           add_timer(0) { signal_loopbreak }
         end
         @reactor_thread = Thread.current
-        run_machine
+
+        # Rubinius needs to come back into "Ruby space" for GC to work,
+        # so we'll crank the machine here.
+        if defined?(RUBY_ENGINE) && RUBY_ENGINE == "rbx"
+          while run_machine_once; end
```
evanphx: Just the tiniest of style notes, I like to use `true while run_machine_once` for this kind of loop.

sodabrew: Ooh, I like that. Next time!

Issue #401, opened 2013-02-08
Assertion Failure in _RunKQueueOnce while using TweetStream Gem with OSX


AtlaStar:
received this message randomly while using an eventmachine reactor wrapped in the TweetStream Gem

Assertion Failed!: (errno != EINVAL), function _RunKQueueOnce, file em.cpp, line 608

Happened at random while maintaining an open public stream with Twitter and figure since TweetStream's eventmachine dependency is built from this repository that I should post the issue here.

The program uses Ruby 1.8.7 that shipped with the OS, and uses gems built with rubygems 1.8.2

My question is if it is currently understood why OSX causes the assertion message at what appears to be random intervals, and since assert() calls abort on the program when it fires, if there is anyway to intercept the value of kqueue inside of the ruby program to throw the kqueue data to a catch block if kqueue's data is consistent with an error code before em.cpp fires the assert()

Issue #372, opened 2012-10-02
EventMachine crashes after the next error


drusakov:
EventMachine craches after the next error:
Assertion failed: (nbytes > 0), function _WriteOutboundData, file ed.cpp, line 1024.

Issue #176, opened 2011-03-08
Assertion failed: (OutboundPages.size() > 0), function Write, file ed.cpp


rb2k:
With the current beta 3 and the latest stable, I receive these messages using em-http-request, em-dns and em-redis:

    Assertion failed: (OutboundPages.size() > 0), function Write, file ed.cpp, line 1563.
    Abort trap

I'd be happy to provide more information if I knew what would be helpful :)
OSX + kqueue and I raised the filedescriptor limit to 2048 (ulimit -n 2048)

PR #619, opened 2015-08-03
Postpone ArmKqueueWriter until all events are processed

### LINKED ISSUE #51: Using kqueue could cause assertion failed
Script: http://gist.github.com/257629
Output: http://gist.github.com/257628#file_output

Basically it nails down to this: it has 2 socket descriptors, kevent returns both. While processing first EM arms kqueue for writing, then sends all the data. And in the next kevent tick it fails assertion.

Reproducible with 0.12.10 on both freebsd and mac os x.

I come up with this patch: https://gist.github.com/bced7e3402d8aa482681

I add bKqueueArmWrite, and do actual arming in _ModifyDescriptors, so all data are sent to sockets at this point.

sodabrew:
This PR implements the patch from #51. Tested on Mac OS X 10.10.4, where the original bug still occurs and the proposed patch resolves it.

Resolves #51, #176, #372, #401.

### DIFF
--- a/ext/ed.cpp
+++ b/ext/ed.cpp
@@ -466,8 +466,9 @@ void ConnectionDescriptor::_UpdateEvents(bool read, bool write)
 	#ifdef HAVE_KQUEUE
 	if (read && SelectForRead())
 		MyEventMachine->ArmKqueueReader (this);
-	if (write && SelectForWrite())
-		MyEventMachine->ArmKqueueWriter (this);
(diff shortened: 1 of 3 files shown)


tmm1:
👍 thanks

sodabrew:
Thanks! Ideally I'd like to turn the triggering script into a minimal repro test case before merge, but only if I can manage it pretty soon.

PR #627, opened 2015-08-11
Use rake-compiler-dock to cross-compile gems for Windows

sodabrew:
With thanks again for the great tools to @kou and @larskanis.

### DIFF
--- a/eventmachine.gemspec
+++ b/eventmachine.gemspec
@@ -1,26 +1,26 @@
-# -*- encoding: utf-8 -*-
-$:.unshift File.expand_path("../lib", __FILE__)
-require "em/version"
-
+require File.expand_path('../lib/em/version', __FILE__)
 
 Gem::Specification.new do |s|
   s.name = 'eventmachine'
   s.version = EventMachine::VERSION
   s.homepage = 'http://rubyeventmachine.com'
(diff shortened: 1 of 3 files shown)


larskanis:
OpenSSL seems to be disabled, with 'rake gem:windows' for now. Tiny_tds has
a shared library based MiniPortile recipe for OpenSSL [here](
https://github.com/rails-sqlserver/tiny_tds/blob/master/ext/tiny_tds/extconf.rb)
and there is the static one in pg.

Everything else LGTM. Unfortunately I don't have a computer this week to do
any tests.

sodabrew:
I've been hoping that RubyInstaller would begin shipping OpenSSL headers for those compiling EventMachine with RubyInstaller + DevKit. But this won't help for binary gems, so maybe I should follow their recipe for static linking. Thanks for the pointer!

larskanis:
> I've been hoping that RubyInstaller would begin shipping OpenSSL headers
for those compiling EventMachine with RubyInstaller + DevKit. But this
won't help for binary gems,

In fact, it would help! If OpenSSL headers are part of the DevKit, I'll add
it to rake-compiler-dock as well, so that it links dynamically to
RubyInstaller's OpenSSL-DLLs at runtime.

PR #630, opened 2015-08-21
Switch from gethostbyname to getaddrinfo, support IPv6 addresses on Windows

### LINKED ISSUE #595: Error on start_server with ipv6 under Windows
If I try to start a server passing "::1" as the host to bind to I get an error.

```ruby
EventMachine.run {
  EventMachine.start_server "::1", 8081, EchoServer
}
```
throws

```
C:/Ruby/lib/ruby/gems/2.2.0/gems/eventmachine-1.0.7/lib/eventmachine.rb:526:in `start_tcp_server': no acceptor (port
 is in use or requires root privileges) (RuntimeError)
        from C:/Ruby/lib/ruby/gems/2.2.0/gems/eventmachine-1.0.7/lib/eventmachine.rb:526:in `start_server'
        from ./echoserver2.rb:20:in `block in <main>'
        from C:/Ruby/lib/ruby/gems/2.2.0/gems/eventmachine-1.0.7/lib/eventmachine.rb:187:in `call'
        from C:/Ruby/lib/ruby/gems/2.2.0/gems/eventmachine-1.0.7/lib/eventmachine.rb:187:in `run_machine'
        from C:/Ruby/lib/ruby/gems/2.2.0/gems/eventmachine-1.0.7/lib/eventmachine.rb:187:in `run'
```

sodabrew:
Resolves #595 by merging and improving upon ibc/eventmachine-le#27 and ibc/eventmachine-le#28
With thanks to @mpalmer for the getaddrinfo implementation in eventmachine-le.

### DIFF
--- a/ext/ed.cpp
+++ b/ext/ed.cpp
@@ -1725,7 +1725,8 @@ void DatagramDescriptor::Write()
 		OutboundPage *op = &(OutboundPages[0]);
 
 		// The nasty cast to (char*) is needed because Windows is brain-dead.
-		int s = sendto (sd, (char*)op->Buffer, op->Length, 0, (struct sockaddr*)&(op->From), sizeof(op->From));
+		int s = sendto (sd, (char*)op->Buffer, op->Length, 0, (struct sockaddr*)&(op->From),
+		               (op->From.sin6_family == AF_INET6 ? sizeof (struct sockaddr_in6) : sizeof (struct sockaddr_in)));
 #ifdef OS_WIN32
 		int e = WSAGetLastError();
 #else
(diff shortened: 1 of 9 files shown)


sodabrew:
So, it turns out in the docs for `getaddrinfo`, that if it detects a literal IPv4 or IPv6 address, it'll just copy that into the result `addrinfo` structure. That may mean we can ditch the inet_pton and the WSAStringToAddress code I just wrote entirely, and rely on getaddrinfo for the whole shebang.

sodabrew:
The WSAStringToAddress commit is now in #632. That'll be safe to land for 1.0.9. Meanwhile, switching to entirely to getaddrinfo will be great for the 1.2.

PR #637, opened 2015-09-28
Replace Exception class with StandardError

shaicoleman:
Don't raise/rescue Exception classes, use StandardError instead.

https://robots.thoughtbot.com/rescue-standarderror-not-exception
https://www.relishapp.com/womply/ruby-style-guide/docs/exceptions
http://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby/10048406#10048406


sodabrew:
I re-ran the failing Ruby 2.2 test, unrelated random failure. Nice work here!

sodabrew:
This is definitely better than directly raising Exception, but I think the best pattern would be a new error class, like EventMachine::Error, and then raise that class or subclasses of it for specific problems.

PR #639, opened 2015-09-29
Support proc-sources in EM::Iterator

jdelStrother:
eg `EM::Iterator.new(proc{ queue.pop || EM::Iterator::Stop })`

Would welcome any comments on this.  The `EM::Iterator::Stop` is kinda weird - a simpler API might be to just return `nil`, but that would assume that no-one's interested in processing nils in their iterator.

### DIFF
--- a/lib/em/iterator.rb
+++ b/lib/em/iterator.rb
@@ -41,6 +41,7 @@ module EventMachine
   #   end
   #
   class Iterator
+    Stop = "EM::Stop"
     # Create a new parallel async iterator with specified concurrency.
     #
     #   i = EM::Iterator.new(1..100, 10)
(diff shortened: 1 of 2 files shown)


tmm1:
Cool! Looks like ruby 1.8 has a `Proc#to_a` which is causing failures there:

```
  test_default_concurrency_with_a_proc:			/home/travis/build/eventmachine/eventmachine/lib/em/iterator.rb:59: warning: default `to_a' will be obsolete
F
===============================================================================
Failure:
test_default_concurrency_with_a_proc(TestIterator)
/home/travis/build/eventmachine/eventmachine/tests/test_iterator.rb:36:in `test_default_concurrency_with_a_proc'
     33:         EM::Timer.new(1) {iter.next}
     34:       }, proc {EM.stop})
     35:     }
  => 36:     assert_equal(10, items.keys.size)
     37:     assert_equal(original_list.to_a.sort, items.values.flatten.sort)
     38:   end
     39: 
<10> expected but was
<1>
diff:
? 10
===============================================================================

```

sodabrew:
Well that was a bad idea: http://ruby-doc.org/core-1.8.7/Object.html#method-i-to_a

jdelStrother:
I changed it so that it checks `respond_to?(:call)` first, which should fix that error.

Issue #186, opened 2011-04-20
unable to install "eventmachine" both from gem and build


suryart:
I am trying to install eventmachine by command: gem install eventmachine and it is giving me this error at the end:

make: shared: Command not found
make: [rubyeventmachine.so] Error 127 (ignored)
/usr/bin/install -c -m 0755 rubyeventmachine.so /usr/local/lib/ruby/gems/1.8/gems/eventmachine-0.12.10/lib
/usr/bin/install: cannot stat `rubyeventmachine.so': No such file or directory
make: **\* [/usr/local/lib/ruby/gems/1.8/gems/eventmachine-0.12.10/lib/rubyeventmachine.so] Error 1

you may review whole log here: http://pastie.org/private/ebiehjoa23klhec6hurmg

And when I tried to grab and build like this:

git clone git://github.com/eventmachine/eventmachine.git
  cd eventmachine
  rake build

at rake build it is prompting following error:

surya@surya-desktop:~/Desktop/Surya/Projects/eventmachine$ rake build
(in /home/surya/Desktop/Surya/Projects/eventmachine)
rake aborted!
Don't know how to build task 'build'

(See full trace by running task with --trace)
surya@surya-desktop:~/Desktop/Surya/Projects/eventmachine$ 

Can you kindly suggest me any possible way to solve this issue?

Thanks
Surya

ScarletMcLearn:
Bump

PR #647, opened 2015-10-17
Add CONFIG to both ext.conf file

YusufAliSiddiqui:
Solves the problem of not being able to install eventmachine.

As seen here, https://github.com/eventmachine/eventmachine/issues/186


sodabrew:
Thank you, I'll review this.
(Please remove the extra ~ swap files however!)

YusufAliSiddiqui:
My bad. Removed the swap files.

sodabrew:
Ok, I see - so this is not a correct fix in general because g++ is not the only compiler, and if there was a prior value for CONFIG["CXX"] we should not overwrite that. Ruby is supposed to, and does, provide CONFIG["CXX"] for us.

The workaround seemed to be needed only for the specific user who had a strange misconfiguration.

Are you seeing this problem yourself now?

YusufAliSiddiqui:
Alright. 

I did have a similar problem myself, and the solution worked for me as well. Perhaps I had the same misconfiguration. 

That's why I thought it might be an apt change.

sodabrew:
Interesting! What OS and Ruby version are you running on? If you run `irb` what do you see for `RbConfig::CONFIG["CXX"]`, for example, on my OS X machine, I get:

```
$ irb
irb(main):003:0> RbConfig::CONFIG["CC"]
=> "xcrun clang"
irb(main):004:0> RbConfig::CONFIG["CXX"]
=> "xcrun clang++"
```

YusufAliSiddiqui:
Hmm. When I run the commands I get gcc and g++ respectively. 
I'm using Ubuntu 14.04 and ruby 2.2.1

sodabrew:
This suggests you do not need any workaround. What error appears when you install the eventmachine gem?

YusufAliSiddiqui:
Actually, the gem installs fine. When I run bundle install on the this gem file, http://pastie.org/10491556

I get the following error. http://pastie.org/10491553

sodabrew:
You will need eventmachine 1.0.4 or higher to compile with Ruby 2.2.x, you've got 1.0.3 which isn't updated for the Ruby API changes.

YusufAliSiddiqui:
I do have 1.0.8, don't know why it keeps using 1.0.3. Any suggestions?

sodabrew:
Is this inside of a project with a Gemfile?

YusufAliSiddiqui:
Yes it is.

sodabrew:
I meant if you have a Gemfile.lock, or if another dependency has a specific version of eventmachine in the Gemfile. (note that the typical use of a Gemfile excludes all other systemwide gems.)

YusufAliSiddiqui:
Yeah, I do have a gemfile.lock .

Okay, so I think I figured it out. I changed the version of the eventmachine in the .lock file, and now it doesn't give that error. 

However, it's giving "An error occurred while installing pg (0.17.1), and Bundler cannot continue." this now.

YusufAliSiddiqui:
Turns out, I had to install postgresQL too. It works now. Thanks!

PR #654, opened 2015-11-02
Support for choosing SSL/TLS versions and ciphers

sodabrew:
This is probably the most substantial port / update of functionality from EventMachine LE.

The usage for `start_tls(:protocols => [TLSv1, ...])` is different than EM-LE, and more like Apache and nginx. The usage for `start_tls(:cipher_list => String)` is identical to EM-LE.

Supersedes and resolve a slew of issues, i'll update this with links.


timon:
Hi!

I'm trying this out because our partner drops support for SSL and migrating to TLS-only connections.

I've added `:protocols => ["tlsv1"]` to my code, but issuing 10 requests in a row results in 7 TLS handshakes and 3 SSL, and I don't have any idea why.

sodabrew:
Thanks for testing this branch! (Just to confirm, you've built the gem from the branch revision? This feature is not yet in any released version of eventmachine.)

timon:
I'm using master from github after the request got merged
13 нояб. 2015 г. 17:56 пользователь "Aaron Stone" <notifications@github.com>
написал:

sodabrew:
Oh right, yes it's landed on master. Do you have debugging information
about where the TLS connection may have failed?

timon:
Sorry for late post.

Our firewall drops connection (by sending RST packet) when it sees forbidden TLS traffic, so OS returns `Errno::ECONNRESET`

PR #662, opened 2015-12-03
Allow reusing of datagram socket/setting bind device

verxx:
- create new function open_datagram_socket_with_reuseaddr
  - the SO_REUSEADDR flag needs to be set before we call the bind on the socket
- add the support to be able to set_sock_opt with SO_BINDTODEVICE
  - this takes in a struct with an interface name set inside of it

### DIFF
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -1644,6 +1644,12 @@ const uintptr_t EventMachine_t::OpenDatagramSocket (const char *address, int por
 	if (sd == INVALID_SOCKET)
 		goto fail;
 
+	{ // set the SO_REUSEADDR on the socket before we bind, otherwise it won't work for a second one
+		int oval = 1;
+		if (setsockopt (sd, SOL_SOCKET, SO_REUSEADDR, (char*)&oval, sizeof(oval)) < 0)
+			goto fail;
+	}
+
 	// Set the new socket nonblocking.
 	if (!SetSocketNonblocking (sd))
 		goto fail;


sodabrew:
This looks pretty good! Could you redo the spacing on the C/C++ source files using tabs instead of spaces, in order to match the code already there?

ext/rubymain.cpp
```diff
@@ -744,24 +744,33 @@ static VALUE t_set_sock_opt (VALUE self UNUSED, VALUE signature, VALUE lev, VALU
...
-		StringValue(optval);
-		v = RSTRING_PTR(optval);
-		len = RSTRING_LENINT(optval);
-		break;
-	}
+  if (option == SO_BINDTODEVICE) {
+    struct ifreq interface; // create struct
+    StringValue(optval); // parse the ruby string
+
+    strncpy(interface.ifr_name, RSTRING_PTR(optval), IFNAMSIZ); // copy interface to struct
```
sodabrew: This should be `StringValueCStr(optval)` to ensure that it is nul-terminated. The RSTRING_PTR can point to a Ruby value that doesn't have a nul-terminator; you need RSTRING_LEN to read only the number of valid bytes. And since `strncpy` does not guarantee a nul-terminator, if the Ruby string doesn't provide one then you may get quite a bit of garbage in there!

sodabrew:
I might suggest that the REUSEADDR should go directly into the existing `open_datagram_socket` function, since the corresponding TCP function already sets SO_REUSEADDR.

verxx:
I didn't include the RESUSEADDR into the existing `open_datagram_socket` function because it alters the function parameters and causes one of the tests to fail because it relies on the last `args` parameter. I'm not sure if changing this will break any other projects or not.

I will address the other comments this week.

sodabrew:
I wouldn't even make it configurable, this should be the behavior in order to remain consistent with the TCP version of the function. I would merge this for the 1.2.0 release so that it won't be a surprise 1.0.x change.

verxx:
Any thing else I need to fix?

sodabrew:
Thanks for the ping! I hadn't checked back to see the updated commits. Could you squash the updates into the original commit?

verxx:
Ok. Done! Thanks for replying. Let me know if there's anything else.

ext/project.h
```diff
@@ -67,6 +69,10 @@ typedef int SOCKET;
 #endif
 #endif /* OS_SOLARIS8 */
 
+#ifndef SO_BINDTODEVICE
+#define SO_BINDTODEVICE 25
+#endif
```
sodabrew: I did some reading on this, and it's a linux-only option. I don't think we should define the value if it is not present; that's an indication that the platform does not support it.

ext/rubymain.cpp
```diff
@@ -744,23 +744,30 @@ static VALUE t_set_sock_opt (VALUE self UNUSED, VALUE signature, VALUE lev, VALU
...
-		i = 1;
-		numval:
-		v = (void*)&i; len = sizeof(i);
-		break;
-	default:
-		StringValue(optval);
-		v = RSTRING_PTR(optval);
-		len = RSTRING_LENINT(optval);
-		break;
+	if (option == SO_BINDTODEVICE) {
```
sodabrew: Since this is linux-only, I now thing we should wrap it in an `#ifdef SO_BINDTODEVICE`. For the consumer code, it looks like Ruby provides this constant since 1.8.7, so we're good for it being exposed:
http://rxr.whitequark.org/mri/source/ext/socket/socket.c?v=1.8.7-p374#4583

verxx:
Yup. You're right. Should be fixed now.

sodabrew:
This looks really good. I definitely want to merge the SO_REUSEADDR code.

However, it turns out you may not need any special code to support SO_BINDTODEVICE after all... the Ruby documentation for `setsockopt` shows how to generate a packed struct here:
http://ruby-doc.org/stdlib-1.8.7/libdoc/socket/rdoc/BasicSocket.html#method-i-setsockopt-label-Parameters
and looking at the `struct ifreq` which has the interface char array right at the beginning:
https://bytes.com/topic/python/answers/30563-passing-argument-setsockopt-socket-sol_socket-so_bindtodevice-intr
and
http://stackoverflow.com/questions/9127311/how-can-you-choose-the-network-interface-via-which-udp-broadcast-packet-is-sent

you should be able to do the device binding like this (note the explicit nul byte in the Ruby string):

``` ruby
EM.run do
  EM.connect "example.com", 80, Module.new {
    define_method :post_init do
      set_sock_opt Socket::SOL_SOCKET, Socket::SO_BINDTODEVICE, "eth0\0"
      EM.stop
    end
  }
end
```

I tried this with "eth0\0" and got a successful response from set_sock_opt,  and with "eth4\0" on a system that had no eth4, and got an appropriate error: `lib/em/connection.rb:280:in `set_sock_opt`: No such device - setsockopt (Errno::ENODEV)`

Granted that it is slightly gross to fake the struct, it is consistent with Ruby's own `setsockopt`. Actually I think I could replace the EM implementation of set_sock_opt with a wrapper around Ruby's own `setsockopt` on the Ruby side and remove this C function entirely!

verxx:
Ok I verified that it works. New change is ready. I think I was using 1.0.0 which I believe was not working for this correctly.

Btw, I would agree with your wrapper idea. That way the code is not duplicated.

ext/project.h
```diff
@@ -43,12 +43,14 @@ See the file COPYING for complete licensing information.
...
 #include <sys/resource.h>
 #include <sys/wait.h>
+//#include <sys/ioctl.h>
 #include <assert.h>
 #include <unistd.h>
 #include <fcntl.h>
 #include <errno.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
+//#include <net/if.h>
```
sodabrew: Can remove the commented lines too :)

verxx:
Fixed :) Thanks!

PR #655, opened 2015-11-04
Add flag SSL_VERIFY_FAIL_IF_NO_PEER_CERT

sodabrew:
Add SSL_VERIFY_FAIL_IF_NO_PEER_CERT to resolve #299, #275, #195.

Add options for `start_tls(:ca_file..., :ca_path..., :private_key_pass...)` to resolve #397, #378, #334, #84.

Interesting discussion about needing extra depth information for `ssl_verify_peer` at
https://github.com/eventmachine/eventmachine/pull/378/files#r2063643 - I wonder if there's a reasonable way to detect the arity of the callback before passing the extra arguments.

Issue #275, opened 2011-11-02
ssl_verify_peer() not called on CA mismatch of cert and key


Zapotek:
Code: https://gist.github.com/1332896
Link to the PEM files required for it are found in the comment.

If the server is run as is, it produces the desirable output of:

```
client: Sending: Hi
client: VERIFY
client: VERIFY
server: VERIFY
server: VERIFY
server: VERIFY
server: Received: Hi 
```

However, if you alter it like so (so that the key and cert are from different CAs):

```
client_opts = {
    :ssl => {
        # :private_key_file => 'key.pem',
        :private_key_file => 'foo-key.pem',
        :cert_chain_file  => 'cert.pem',
        # :cert_chain_file  => 'foo-cert.pem',
    },
    :role => :client
} 
```

The script produces:

```
client: Sending: Hi
client: VERIFY
client: VERIFY
server: Received: Hi
```

Which means that `ssl_verify_peer()` is being bypassed and the request is being forwarded to `receive_data()` which should _not_ have happened as it nullifies any sort of authN/authZ system that relies on SSL peer verification.

Issue #195, opened 2011-05-02
ssl_verify_peer() when the client provides no cert (proposal)


ibc:
If a client provides no certificate in the TLS negociation `ssl_verify_peer(cert)` is not called (as such method is called for each certificate provided by the client).

Let's say I want to deny a connection if the client provides an invalid certificate, but also if the client provides no certificate.
How to handle this last case? A workaround would be using `cert=get_peer_cert()` in `ssl_handshake_completed()` method and deny the connection if `cert` is nil.

However, I suggest than when the EM server runs `start_tls` with `:verify_peer => true`, the `ssl_verify_peer(cert)` is also called even if the client no provides a certificate. In this case, the method would be called with `nil` as argument. In this way the server realizes that no certificate is provided so can reject the connection.

Another solution could be a new callback `no_peer_ssl()` which is invoked in case no certificate is provided by client. Honestly I prefer the solution above (simpler IMHO).

PR #665, opened 2015-12-13
Add start_tls opts :ecdh_curve, :dhparam and :fail_if_no_peer_cert

mwchambers:


sodabrew:
Thanks for the very thorough PR! Could you also describe the purpose of the change and help me with some background information on the purpose of these new DH Curve and ECDH Curve options?

I see https://wiki.openssl.org/index.php/Elliptic_Curve_Diffie_Hellman but at first glance I did not understand what the end-user could/should do to configure these options.

Adding VERIFY_FAIL_IF_NO_PEER_CERT is already in #655, btw - which is intended to be a larger roll-up of prior SSL feature requests, and would set the stage for adding more detailed settings like the ones you're proposing here.

mwchambers:
Hi,

I had seen the other commit which adds VERIFY_FAIL_IF_NO_PEER_CERT to the options. However I wanted more flexibility and figured that adding it as an additional param may also prevent changing the behaviour of existing applications built on top of event machine.

I was looking at using thin to build a rack app which would work without peer certificates but behave differently if they were available. If you look at ssl_verify_peer and pre_process methods inside https://github.com/macournoyer/thin/blob/master/lib/thin/connection.rb it sets rack.peer_cert in the env which will let me do this. By always setting VERIFY_FAIL_IF_NO_PEER_CERT this would not be possible. OTOH I have a separate application which needs to reject all clients without a peer cert, so failing in the SSL handshake makes more sense here.

The :ecdh_curve and :dhparam options are to get openssl to enable the 'perfect forward security' cipher suites. Without these options being provided those cipher suites are unavailable. For example if you connect Chrome to an eventmachine server and examine the connection details you will probably see it complaining that the connecton is 'using an obsolete cipher suite', unhappy about the key exchange mechanism. I've added both options so both ECDHE and DHE are available so client's using Internet Explorer can use AES GCM based ciphers ( https://tools.ietf.org/html/rfc5288 ). The version of Internet explorer didn't support ECDHE.

The :dhparam option needs to just be set to the path of a file which should be generated with 'openssl dhparam'.

The :edch_curve option is a pre-defined curve name defined by openssl. You can list then using 'openssl ecparam -list_curves'

Read more here:

https://wiki.openssl.org/index.php/Diffie-Hellman_parameters
https://wiki.openssl.org/index.php/Elliptic_Curve_Cryptography
https://wiki.openssl.org/index.php/Elliptic_Curve_Diffie_Hellman

You could make eventmachine set defaults if these options are not set. However I'm unsure what would be a sensible default for the named curve.

The openssl wiki above shows examples of hard coding the dhparam into the source code. But doing this seems to contradict the recommendations in the documentation: 

"The risk in reusing DH parameters is that an attacker may specialize on a very often used DH group. Applications should therefore generate their own DH parameters during the installation process using the openssl dhparam application."

from: https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_tmp_dh_callback.html. 

Hopefully someone with a better understanding of OpenSSL could provide sane defaults for these parameters.

sodabrew:
Resolve #299, #275, #195.

PR #690, opened 2016-02-27
Fix build on OpenBSD

t0str:
If EPROTO does not exist define it as EINTR

OpenBSD does not have EPROTO: https://sourceforge.net/p/sshguard/mailman/message/21247557/

Works, tried on 1.2.dev.2 and 1.0.9.1.

### DIFF
--- a/ext/ed.cpp
+++ b/ext/ed.cpp
@@ -878,6 +878,9 @@ void ConnectionDescriptor::_DispatchInboundData (const char *buffer, unsigned lo
 
 		// If our SSL handshake had a problem, shut down the connection.
 		if (s == -2) {
+			#ifndef EPROTO // OpenBSD does not have EPROTO
+			#define EPROTO EINTR
+			#endif
 			#ifdef OS_UNIX
 			UnbindReasonCode = EPROTO;
 			#endif


sodabrew:
Looks good! Would you mind adjusting indentation from spaces to tabs to match the surrounding code?

t0str:
Ah yes, did not see that. Should be fine now. Have been running the change for a week and no problems so far (on 1.0.9.1).

dt1973:
Wait, so how'd one go about installing EventMachine on OpenBSD? Tried `gem install eventmachine -v '1.2.0.dev.2'` but getting the same error.

dt1973:
Nevermind, found http://stackoverflow.com/questions/2577346/how-to-install-gem-from-github-source and seems I can do `gem specific_install`.

sodabrew:
Yep, that'll tide you over!

sodabrew:
EventMachine 1.2.0 is now released!

rollcat:
Hi guys, any chance of backporting the fix to the 1.0.9.* branch? Currently [Sensu](https://github.com/sensu/sensu) 0.22.2 still requires eventmachine 1.0.9.1 - it doesn't seem to work with 1.2, but this simple patch against the tag `v1.0.9.1` makes it work perfectly:

```
diff --git a/ext/ed.cpp b/ext/ed.cpp
index a3b23a4..12a4a20 100644
--- a/ext/ed.cpp
+++ b/ext/ed.cpp
@@ -874,6 +874,9 @@ void ConnectionDescriptor::_DispatchInboundData (const char *buffer, unsigned lo
                // If our SSL handshake had a problem, shut down the connection.
                if (s == -2) {
                        #ifdef OS_UNIX
+                       #ifndef EPROTO
+                       #define EPROTO EINTR
+                       #endif
                        UnbindReasonCode = EPROTO;
                        #endif
                        #ifdef OS_WIN32
```

Would it be possible to do another release in the 1.0.9 branch? I have created [an issue with Sensu](https://github.com/sensu/sensu/issues/1194), asking if they'd bump their requirement.

sodabrew:
Definitely, I'll work on it this week.

rollcat:
Sensu has moved to eventmachine 1.2 in a new release, I believe we're good here.

PR #693, opened 2016-03-10
Fix compilation on AIX (w/ XLC)

portertech:
We had to replace `-shared` with `-bstatic` + `-bdynamic` in order to get EM (>= 1.0.9.1) to compile/build/install on AIX, using the XLC compiler. AFAIK GCC should never be used.

### DIFF
--- a/ext/extconf.rb
+++ b/ext/extconf.rb
@@ -211,7 +211,7 @@ when /linux/
   CONFIG['LDSHARED'] = "$(CXX) -shared"
 
 when /aix/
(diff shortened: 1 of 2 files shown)


portertech:
`Gem::InstallError: rake requires Ruby version >= 1.9.3.` :-(

sodabrew:
The comments suggest they were intended for gcc -- should there be different flags for compiling on AIX with XLC vs. GCC ?

sodabrew:
Oh, wow Rake 11. ok.

PR #694, opened 2016-03-11
Close socket on close_connection even after close_connection_after_writing

mdpye:
Previously, calls to close now were ignored if a close was scheduled for after we finish writing. This was problematic in any case where a graceful close was scheduled, but the send buffer might never be consumed, for example:
 - stuck TCP connections
 - extremely slow readers


sodabrew:
Thank you!

PR #701, opened 2016-03-16
Improved build

thedrow:
Faster build times for event machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,12 @@ before_install:
 env:
   global:
     - TESTOPTS=-v
+    - MAKE="make -j 2"
 language: ruby
 sudo: false
+cache:
+  - bundler
+  - ccache
 rvm:
   - 2.3.0
   - 2.2


thedrow:
Build failures do not seem related to me.

sodabrew:
Thanks! It looks like it's only a few second improved, though? I'll try re-running the builds to see if the cache kicks in.

I'm also interested in switching to the Trusty images, but that can be another day's PR :)

The failures are due to Rake 11 not supporting Ruby < 1.9.3 anymore, but rubygems is still trying to install it instead of the last version that did support Ruby >= 1.8.7. I need to figure out what I'll do with this to make the builds succeed. I have not yet dropped Ruby 1.8.7 support from EventMachine.

sodabrew:
Yeah... it's still between 1:20 and 1:45 for the builds, with or without `make -j 2` or the caches. Without much change in running time, I would prefer to keep the travis config simpler.

thedrow:
I don't see that the cache bucket for ccache is added.
That's probably why.
make -j 2 should also help because Travis has two cores. I wonder why it doesn't.

PR #706, opened 2016-03-28
Add regular expression (regex) support to LineText2 protocol's delimiter

taranda:
This mod allows the use of a regular expression as the end-of-line delimiter in the LineText2 protocol.  The set_delimiter method will accept a regular expression and set the end-of-line delimiter appropriately.  Anything passed to the set_delimiter method other than a regular expression will be converted to a string.  A new test to capture the use of a regular express was added.

### DIFF
--- a/lib/em/protocols/linetext2.rb
+++ b/lib/em/protocols/linetext2.rb
@@ -56,7 +56,14 @@ module EventMachine
 
         while remaining_data.length > 0
           if @lt2_mode == :lines
-            if ix = remaining_data.index( @lt2_delimiter )
+            delimiter_string = case @lt2_delimiter
+            when Regexp
+              remaining_data.slice(@lt2_delimiter)
+            else
+              @lt2_delimiter
+            end
+            ix = remaining_data.index(delimiter_string) if delimiter_string
+            if ix
               @lt2_linebuffer << remaining_data[0...ix]
               ln = @lt2_linebuffer.join
               @lt2_linebuffer.clear
@@ -64,7 +71,7 @@ module EventMachine
                 ln.chomp!
               end
               receive_line ln
-              remaining_data = remaining_data[(ix+@lt2_delimiter.length)..-1]
+              remaining_data = remaining_data[(ix+delimiter_string.length)..-1]
             else
               @lt2_linebuffer << remaining_data
               remaining_data = ""
@@ -101,9 +108,16 @@ module EventMachine
         end
       end
 
-
+      # The line delimiter may be a regular expression or a string.  Anything
+      # passed to set_delimiter other than a regular expression will be
+      # converted to a string.
       def set_delimiter delim
-        @lt2_delimiter = delim.to_s
+        @lt2_delimiter = case delim
+        when Regexp
+          delim
+        else
+          delim.to_s
+        end
       end
 
       # Called internally but also exposed to user code, for the case in which
@@ -163,4 +177,3 @@ module EventMachine
     end
   end
 end
-
--- a/tests/test_ltp2.rb
+++ b/tests/test_ltp2.rb
@@ -72,6 +72,30 @@ class TestLineText2 < Test::Unit::TestCase
     assert_equal( ["Linea", "Lineb", "Linec", "Lined"], a.lines )
   end
 
+  class RegexDelimiter
+    include EM::Protocols::LineText2
+    attr_reader :lines
+    def initialize *args
+      super
+      @delim = /[A-D]/
+      set_delimiter @delim
+    end
+    def receive_line line
+      (@lines ||= []) << line
+    end
+  end
+
+  def test_regex_delimiter
+    testdata = %Q(LineaALinebBLinecCLinedD)
+
+    a = RegexDelimiter.new
+    a.receive_data testdata
+    assert_equal( ["Linea", "Lineb", "Linec", "Lined"], a.lines )
+
+    a = RegexDelimiter.new
+    testdata.length.times {|i| a.receive_data( testdata[i...i+1] ) }
+    assert_equal( ["Linea", "Lineb", "Linec", "Lined"], a.lines )
+  end
 
   #--
   # Test two lines followed by an empty line, ten bytes of binary data, then


lib/em/protocols/linetext2.rb
```diff
@@ -101,9 +102,15 @@ def receive_data data
         end
       end
 
-
+      # The line delimiter may be a regular expression or a string.  Anything 
+      # passed to set_delimiter other than a regular expression will be 
+      # converted to a string.
       def set_delimiter delim
-        @lt2_delimiter = delim.to_s
+        if delim.class == Regexp
```
sodabrew: The `case` block may be cleaner:
``` ruby
@lt2_delimiter = case delim
when Regexp
   delim
else
  delim.to_s
end
```

taranda:
Thanks for the recommendation.  I modified the commit with your code recommendation.

taranda:
Somehow I captured thedrow's commit in this pull request.  Let me know and I will delete this pull request and submit a new one.

sodabrew:
Since you're submitting your commit from your master branch - let's assume that you have a remote called upstream - try this:

``` shell
$ git checkout master
$ git fetch upstream
$ git rebase upstream/master
$ git push -f origin master
```

To check on your remotes (this is how mine look):
```shell
$ git remote -v
origin	git@github.com:sodabrew/eventmachine (fetch)
origin	git@github.com:sodabrew/eventmachine (push)
upstream	git@github.com:eventmachine/eventmachine.git (fetch)
upstream	git@github.com:eventmachine/eventmachine.git (push)
```

To add a remote:
``` shell
$ git remote add upstream git@github.com:eventmachine/eventmachine.git
```

sodabrew:
You also do not need to submit new pull requests. A PR matches with a source/dest branch. You can add and remove commits on your branch and the PR will update to reflect whatever is on your branch.

taranda:
I can try reverting and re-commiting.

I am not sure what to do here.  Unfortunately, I do not have much experience with pull requests.

taranda:
I think the cleanest solution may be to close this pull request and submit a new one.  Thoughts?

sodabrew:
I supplied the necessary commands above, happy to help explain further if it's not working.

sodabrew:
Now that you've added a revert commit that we do not want to have on this branch, you'll need to do an interactive rebase:

``` shell
$ git rebase -i upstream/master
```

When you're inside the editor window, delete the line with the README commit and delete the line with your Revert README commit. Leave only the line with your Line Protocol commit. Then save and exit and let git do the work of adjusting the commits on your local branch.

Force-push your adjusted branch to GitHub:

``` shell
$ git push -f origin master
```

taranda:
I reset back to before the revert commits.  I think it is fixed now.  When you merge, you may want to make sure that README.md is not negatively affected.  I did not modify it, but it showed modified on my forked master branch when I reset back to my original commit.  This appears to have worked and thank you for your help.  I use git everyday, but I learned something new today.

sodabrew:
You got it! Well done, let's let the tests execute...

lib/em/protocols/linetext2.rb
```diff
@@ -57,6 +57,7 @@ def receive_data data
         while remaining_data.length > 0
           if @lt2_mode == :lines
             if ix = remaining_data.index( @lt2_delimiter )
```
sodabrew: Does `String#index` behave correctly with a Regex as the delimiter here?

lib/em/protocols/linetext2.rb
```diff
@@ -57,6 +57,7 @@ def receive_data data
...
+              delimiter_length = remaining_data.slice(@lt2_delimiter).length
```
sodabrew: If `remaining_data.index(@l2_delimiter)` above is executing the Regexp, then we probably don't want to execute it a second time to get the length of the match; I'm thinking what an alternate mechanism might look like to execute the Regexp just once per chunk.

taranda: I agree that it is inefficient, but I am not sure how to do this with only one run of the regex.  The code I replaced also had a separate index call and length call.

taranda: We could use slice to extract the matching string and then do to a string search to extract the index.  That will add a line or two of code but only execute the regex match once.  It will probably be more efficient in execution.

`if ix_str = remaining_data.slice( @lt2_delimiter )
   ix = remaining_data.index(ix_str)
   delimiter_length = ix_str.length`

Something like that maybe?

taranda: My markdown was poor...here is the code segment:

`if delimiter_string = remaining_data.slice( @lt2_delimiter )`
`  ix = remaining_data.index(delimiter_string)`
`  delimiter_length = delimiter_string.length`

Actually, we could just get rid of the `delimiter_length` variable and put `delimiter_string.length` in the appropriate place lower down.  I'll do that and update.

lib/em/protocols/linetext2.rb
taranda: It does according to ruby's documentation and the test.

taranda:
I have refactored the code.  The new code is not a elegant to read, but I believe it will be much more efficient in execution.  The only additional computation cost for the string case is a handful of extra conditional operations and some string assignments.  There are no extra searches required. 

In the regular expression case, the Regex is only evaluated once and further searches are done with strings which will be much more efficient.  This re-factor has many more lines of code, but I think the improvement in execution speed is worth it.  

Note: I have not done any benchmark testings, but I assume string searches are more expensive than conditionals and string searches.  

Please let me know if you have any questions.

lib/em/protocols/linetext2.rb
```diff
@@ -56,15 +56,22 @@ def receive_data data
...
+            if ix 
               @lt2_linebuffer << remaining_data[0...ix]
               ln = @lt2_linebuffer.join
               @lt2_linebuffer.clear
               if @lt2_delimiter == "\n"
                 ln.chomp!
               end
               receive_line ln
-              remaining_data = remaining_data[(ix+@lt2_delimiter.length)..-1]
+              remaining_data = remaining_data[ix+delimiter_string.length..-1]
```
sodabrew: Suggest adding back the parens for clarity.

lib/em/protocols/linetext2.rb
```diff
@@ -56,15 +56,22 @@ def receive_data data
 
         while remaining_data.length > 0
           if @lt2_mode == :lines
-            if ix = remaining_data.index( @lt2_delimiter )
+            delimiter_string = case @lt2_delimiter
```
sodabrew: Trailing whitespace on this line and a couple of others.

lib/em/protocols/linetext2.rb
```diff
@@ -56,15 +56,22 @@ def receive_data data
...
         while remaining_data.length > 0
           if @lt2_mode == :lines
-            if ix = remaining_data.index( @lt2_delimiter )
+            delimiter_string = case @lt2_delimiter 
+            when Regexp
+              remaining_data.slice(@lt2_delimiter)
+            else
+              @lt2_delimiter
+            end
+            ix = (delimiter_string.nil?) ? nil : remaining_data.index(delimiter_string)
```
sodabrew: This might be better as `ix = remaining_data.index(delimiter_string) if delimiter_string`

taranda:
Thanks for the suggestions.  The pull request has been updated.

PR #708, opened 2016-04-05
fixes compilation on latest ubuntu

schmurfy:
I am not sure why but I need this for the shared library to be properly linked to libstdc++ on ubuntu 15.10.

### DIFF
--- a/ext/extconf.rb
+++ b/ext/extconf.rb
@@ -206,6 +206,7 @@ when /darwin/
 
 when /linux/
   add_define 'HAVE_EPOLL' if have_func('epoll_create', 'sys/epoll.h')
+  check_libs(%w[stdc++], true)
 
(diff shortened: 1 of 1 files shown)


sodabrew:
Interesting that you need this. The change looks reasonable, but I want to make sure that it's not something odd about your Ruby build or missing packages -  is this with the system Ruby? which compiler?

schmurfy:
that's not the 

I know that's weird xD
I will do more tests to check this is not my setup although I don't think I did something special and report here.

schmurfy:
nevermind I found the culprit, I had this in my environment: `CXX=/usr/bin/gcc`, when you properly compile using g++ libstdc++ get linked automatically.

sodabrew:
Thanks for chasing down the issue and reporting it back here! Glad you're up and running.

PR #712, opened 2016-04-16
Pure Ruby EM SSL (working start_tls)

portertech:
This pull request adds SSL/TLS support to the pure Ruby EM implementation.

@amdprophet & I went into the belly of the beast that is em/pure_ruby. The pure Ruby implementation leaves much to be desired, but it's a start!

```
$ rake test_em_pure_ruby                                         
EM Library Type: pure_ruby
Loaded suite /home/portertech/.gem/ruby/2.3.0/gems/rake-11.1.2/lib/rake/rake_test_loader
Started
.........
======================================================================================================================================
omitted. [test_no_dhparam(TestSslDhParam)]
/home/portertech/projects/sensu/eventmachine/tests/test_ssl_dhparam.rb:49:in `test_no_dhparam'
======================================================================================================================================
.............
======================================================================================================================================
omitted. [test_accept_server(TestSslVerify)]
/home/portertech/projects/sensu/eventmachine/tests/test_ssl_verify.rb:101:in `test_accept_server'
======================================================================================================================================
======================================================================================================================================
omitted. [test_deny_server(TestSslVerify)]
/home/portertech/projects/sensu/eventmachine/tests/test_ssl_verify.rb:116:in `test_deny_server'
======================================================================================================================================
.

Finished in 1.25792021 seconds.
--------------------------------------------------------------------------------------------------------------------------------------
26 tests, 57 assertions, 0 failures, 0 errors, 0 pendings, 3 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------
20.67 tests/s, 45.31 assertions/s
```

### DIFF
--- a/lib/em/pure_ruby.rb
+++ b/lib/em/pure_ruby.rb
@@ -33,6 +33,88 @@ require 'forwardable'
 require 'socket'
 require 'fcntl'
 require 'set'
+require 'openssl'
+
+module EventMachine
+  # @private
+  class Error < Exception; end
+  # @private
+  class UnknownTimerFired < RuntimeError; end
+  # @private
+  class Unsupported < RuntimeError; end
+  # @private
+  class ConnectionError < RuntimeError; end
+  # @private
+  class ConnectionNotBound < RuntimeError; end
+
+  # Older versions of Ruby may not provide the SSLErrorWaitReadable
+  # OpenSSL class. Create an error class to act as a "proxy".
+  if defined?(OpenSSL::SSL::SSLErrorWaitReadable)
+    SSLConnectionWaitReadable = OpenSSL::SSL::SSLErrorWaitReadable
+  else
+    SSLConnectionWaitReadable = IO::WaitReadable
+  end
+
+  # Older versions of Ruby may not provide the SSLErrorWaitWritable
+  # OpenSSL class. Create an error class to act as a "proxy".
+  if defined?(OpenSSL::SSL::SSLErrorWaitWritable)
+    SSLConnectionWaitWritable = OpenSSL::SSL::SSLErrorWaitWritable
+  else
+    SSLConnectionWaitWritable = IO::WaitWritable
+  end
+end
+
+module EventMachine
+  class CertificateCreator
+    attr_reader :cert, :key
+
+    def initialize
+      @key = OpenSSL::PKey::RSA.new(1024)
+      public_key = @key.public_key
+      subject = "/C=EventMachine/O=EventMachine/OU=EventMachine/CN=EventMachine"
+      @cert = OpenSSL::X509::Certificate.new
+      @cert.subject = @cert.issuer = OpenSSL::X509::Name.parse(subject)
+      @cert.not_before = Time.now
+      @cert.not_after = Time.now + 365 * 24 * 60 * 60
+      @cert.public_key = public_key
+      @cert.serial = 0x0
+      @cert.version = 2
+      factory = OpenSSL::X509::ExtensionFactory.new
+      factory.subject_certificate = @cert
+      factory.issuer_certificate = @cert
+      @cert.extensions = [
+        factory.create_extension("basicConstraints","CA:TRUE", true),
+        factory.create_extension("subjectKeyIdentifier", "hash")
+      ]
+      @cert.add_extension factory.create_extension("authorityKeyIdentifier", "keyid:always,issuer:always")
+      @cert.sign(@key, OpenSSL::Digest::SHA1.new)
+    end
+  end
+
+  # @private
+  DefaultCertificate = CertificateCreator.new
+
+  # @private
+  DefaultDHKey1024 = OpenSSL::PKey::DH.new <<-_end_of_pem_
+-----BEGIN DH PARAMETERS-----
+MIGHAoGBAJ0lOVy0VIr/JebWn0zDwY2h+rqITFOpdNr6ugsgvkDXuucdcChhYExJ
+AV/ZD2AWPbrTqV76mGRgJg4EddgT1zG0jq3rnFdMj2XzkBYx3BVvfR0Arnby0RHR
+T4h7KZ/2zmjvV+eF8kBUHBJAojUlzxKj4QeO2x20FP9X5xmNUXeDAgEC
+-----END DH PARAMETERS-----
+  _end_of_pem_
+
+  # @private
+  DefaultDHKey2048 = OpenSSL::PKey::DH.new <<-_end_of_pem_
+-----BEGIN DH PARAMETERS-----
+MIIBCAKCAQEA7E6kBrYiyvmKAMzQ7i8WvwVk9Y/+f8S7sCTN712KkK3cqd1jhJDY
+JbrYeNV3kUIKhPxWHhObHKpD1R84UpL+s2b55+iMd6GmL7OYmNIT/FccKhTcveab
+VBmZT86BZKYyf45hUF9FOuUM9xPzuK3Vd8oJQvfYMCd7LPC0taAEljQLR4Edf8E6
+YoaOffgTf5qxiwkjnlVZQc3whgnEt9FpVMvQ9eknyeGB5KHfayAc3+hUAvI3/Cr3
+1bNveX5wInh5GDx1FGhKBZ+s1H+aedudCm7sCgRwv8lKWYGiHzObSma8A86KG+MD
+7Lo5JquQ3DlBodj3IDyPrxIv96lvRPFtAwIBAg==
+-----END DH PARAMETERS-----
+  _end_of_pem_
+end
 
 # @private
 module EventMachine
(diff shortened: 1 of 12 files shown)


lib/em/pure_ruby.rb
```diff
@@ -159,10 +160,73 @@ def set_timer_quantum interval
...
+      tls_parms = @tls_parms[signature]
+      ctx = OpenSSL::SSL::SSLContext.new(tls_parms[:ssl_version])
+      ctx.cert = DefaultCertificate.cert
+      ctx.key = DefaultCertificate.key
+      ctx.cert_store = OpenSSL::X509::Store.new
+      ctx.cert_store.set_default_paths
+      ctx.cert = OpenSSL::X509::Certificate.new(tls_parms[:cert_chain]) if tls_parms[:cert_chain]
+      ctx.key = OpenSSL::PKey::RSA.new(tls_parms[:priv_key]) if tls_parms[:priv_key]
+      verify_mode = OpenSSL::SSL::VERIFY_NONE
+      if tls_parms[:verify_peer] || tls_parms[:verify_peer] == 1
```
sodabrew: Is this redundant because `1` is truthy?

lib/em/pure_ruby.rb
```diff
@@ -228,10 +292,47 @@ def set_pending_connect_timeout sig, tm
...
   class Error < Exception; end
+  # @private
+  class ConnectionError < RuntimeError; end
+  # @private
+  class ConnectionNotBound < RuntimeError; end
+  # @private
+  class UnknownTimerFired < RuntimeError; end
+  # @private
+  class Unsupported < RuntimeError; end
+end
```
sodabrew: RuntimeError is a child of StandardError, so this is good.

In a separate PR, I am working on getting the rest of EM::Error under StandardError, but RuntimeError may be just as well.

lib/em/pure_ruby.rb
portertech: I'll happily remove the extra conditional, wasn't sure if all consuming libraries were following the rules and using a boolean.

lib/em/pure_ruby.rb
portertech: Do you need me to change any of this?

rakelib/test_pure.rake
```diff
@@ -0,0 +1,13 @@
...
+  t.libs << 'tests'
+  t.libs << 'lib'
+  t.pattern = 'tests/**/test_pure*.rb'
+  t.warning = true
+end
+
+task :test_em_pure_ruby do
+  ENV['EM_PURE_RUBY'] = 'true'
+  Rake::Task['test_pure'].execute
+end
```
sodabrew: I wonder if we could get this harness to execute the entire test suite with the Ruby reactor?

tests/test_pure.rb
```diff
@@ -85,4 +85,55 @@ def test_reactor_running
     assert a
   end
 
+  module TLSServer
```
sodabrew: Would it make sense to use the common ssl tests but with the Pure Ruby reactor under the hood, rather than a separate set of tests?

tests/em_test_helper.rb
```diff
@@ -1,8 +1,11 @@
+require 'em/pure_ruby' if ENV['EM_PURE_RUBY']
```
sodabrew: Given this environment variable, we could trigger it either with the test_pure top-level Rake task, or with a `.travis.yml` flag -- in the latter case, the entire Ruby version and platform matrix could be executed twice to ensure wide compatibility.

lib/em/pure_ruby.rb
```diff
@@ -228,10 +292,47 @@ def set_pending_connect_timeout sig, tm
...
+      @cert.extensions = [
+        factory.create_extension("basicConstraints","CA:TRUE", true),
+        factory.create_extension("subjectKeyIdentifier", "hash")
+      ]
+      @cert.add_extension factory.create_extension("authorityKeyIdentifier", "keyid:always,issuer:always")
+      @cert.sign(@key, OpenSSL::Digest::SHA1.new)
+    end
+  end
+
+  DefaultCertificate = CertificateCreator.new
```
sodabrew: I'm not following the circumstances where this would be used?

portertech: Default TLS certificate, as the C++ reactor does something similar. Required for working TLS server when a private key and cert chain is not provided.

tests/em_test_helper.rb
portertech: Yep :+1: Just wanted an easy way to spot test, as I am sure the current pure ruby reactor will fail many of the tests in the suite.

tests/test_pure.rb
portertech: Yes, that's the end goal. Baby steps. I'll try some more of the common tests.

rakelib/test_pure.rake
portertech: Didn't want existing reactor bugs etc to creep into the PR.

portertech:
Looks like we do not want to set context ssl_version, but instead use context options or ciphers to limit (like the C++ reactor).

lib/em/pure_ruby.rb
sodabrew: Let's consider this PR ahead of my other one -- just a heads up that I may scoop these out into a separate lib/em/error.rb as a followup step.

rakelib/test_pure.rake
sodabrew: The test suite is in reasonable shape, so I wouldn't worry about too many false-negative unrelated failures. I'll give this a try on my local machine later today.

lib/em/pure_ruby.rb
sodabrew: Oh, right. That default cert is super yucky, I am looking forward to removing it. This class might make a reasonable replacement for all of the reactors.

lib/em/pure_ruby.rb
```diff
@@ -180,21 +180,22 @@ def set_tls_parms signature, priv_key, cert_chain, verify_peer, fail_if_no_peer_
...
-        (bits & protocols_bitmask) != 0
+      ssl_options = {
+        EM_PROTO_SSLv2 => OpenSSL::SSL::OP_NO_SSLv2,
+        EM_PROTO_SSLv3 => OpenSSL::SSL::OP_NO_SSLv3,
+        EM_PROTO_TLSv1 => OpenSSL::SSL::OP_NO_TLSv1,
+        EM_PROTO_TLSv1_1 => OpenSSL::SSL::OP_NO_TLSv1_1,
+        EM_PROTO_TLSv1_2 => OpenSSL::SSL::OP_NO_TLSv1_2
+      }.inject(0) do |options, (bits, option)|
+        options |= option if (bits & protocols_bitmask) == 0
+        options
```
sodabrew: Yes, this better matches the C++ by looping over the entire array. Note that `ssl_version` might also be a non-array (but I think I coerce it to an array in lib/em/connection.rb, so by the time it gets to the reactor it should always be an array? gotta double check).

This is something I implemented differently in EM 1.2 vs. how Ruby does it, is allowing an array of possible `ssl_version`s. I want to propose this upstream; the way Ruby does it now makes it fairly hard to write "Allow the following SSL/TLS protocol versions: TLS 1.0, 1.1, or 1.2"

rakelib/test_pure.rake
```diff
@@ -3,7 +3,7 @@ require 'rake/testtask'
 Rake::TestTask.new(:test_pure) do |t|
   t.libs << 'tests'
   t.libs << 'lib'
-  t.pattern = 'tests/**/test_pure*.rb'
+  t.test_files = Dir.glob('tests/**/test_pure*.rb') + Dir.glob('tests/**/test_ssl*.rb')
```
sodabrew: Excellent!

lib/em/pure_ruby.rb
portertech: :+1:

portertech:
@sodabrew I think I have taken this as far as I can for the time being. These changes allow Sensu to run on platforms that cannot successfully compile the c++ reactor :+1:

portertech:
@amdprophet it's happening!

portertech:
I made sure to comment on the few SSL test omission lines.

amdprophet:
giphy

sodabrew:
I'll give this another look over and run the test suite myself in a few hours. Assuming I don't find anything else at issue, are you still working on this, or ready for merge?

portertech:
@sodabrew ready for merge :+1:

sodabrew:
The stock OS X Ruby 2.0 may have an older OpenSSL under the hood?

```
  NameError: uninitialized constant OpenSSL::SSL::SSLErrorWaitReadable
```

and I found myself replacing this nice Hash inject with a less elegant pile of ifs, because I don't have the TLSv1_1, TLSv1_2 macros available:

```
-      ssl_options = {
-        EM_PROTO_SSLv2 => OpenSSL::SSL::OP_NO_SSLv2,
-        EM_PROTO_SSLv3 => OpenSSL::SSL::OP_NO_SSLv3,
-        EM_PROTO_TLSv1 => OpenSSL::SSL::OP_NO_TLSv1,
-        EM_PROTO_TLSv1_1 => OpenSSL::SSL::OP_NO_TLSv1_1,
-        EM_PROTO_TLSv1_2 => OpenSSL::SSL::OP_NO_TLSv1_2
-      }.inject(OpenSSL::SSL::OP_ALL) do |options, (bits, option)|
-        options |= option if (bits & protocols_bitmask) == 0
-        options
-      end
+      options = OpenSSL::SSL::OP_ALL
+      # The basic SSL / TLS versions are widely defined across OpenSSL versions
+      options |= OpenSSL::SSL::OP_NO_SSLv2 if defined?(OpenSSL::SSL::OP_NO_SSLv2) && EM_PROTO_SSLv2 & protocols_bitmask == 0
+      options |= OpenSSL::SSL::OP_NO_SSLv3 if defined?(OpenSSL::SSL::OP_NO_SSLv3) && EM_PROTO_SSLv3 & protocols_bitmask == 0
+      options |= OpenSSL::SSL::OP_NO_TLSv1 if defined?(OpenSSL::SSL::OP_NO_TLSv1) && EM_PROTO_TLSv1 & protocols_bitmask == 0
+      # But these TLS 1.x versions are not defined until later OpenSSL versions
+      options |= OpenSSL::SSL::OP_NO_TLSv1_1 if defined?(OpenSSL::SSL::OP_NO_TLSv1_1) && EM_PROTO_TLSv1_1 & protocols_bitmask == 0
+      options |= OpenSSL::SSL::OP_NO_TLSv1_2 if defined?(OpenSSL::SSL::OP_NO_TLSv1_2) && EM_PROTO_TLSv1_2 & protocols_bitmask == 0
```

portertech:
@sodabrew would you like me to apply your patch to this pr?

sodabrew:
Yes please do. I hadn't looked at the context for `SSLErrorWaitReadable` to see if that can be wrapped in an `if defined?` block, but I think that is important to avoid breaking older OpenSSL versions.

portertech:
@sodabrew I'll make the necessary changes :+1: Thanks for catching these.

lib/em/pure_ruby.rb
```diff
@@ -33,6 +33,18 @@
 require 'socket'
 require 'fcntl'
 require 'set'
+require 'openssl'
+
+# Older versions of Ruby & OpenSSL may not provide the
+# SSLErrorWaitReadable and SSLErrorWaitWritable OpenSSL classes.
+# Set them if not defined.
+unless defined?(OpenSSL::SSL::SSLErrorWaitReadable)
+  OpenSSL::SSL::SSLErrorWaitReadable = OpenSSL::SSL::SSLError
```
portertech: @sodabrew WDYT?

lib/em/pure_ruby.rb
```diff
@@ -159,10 +171,127 @@ def set_timer_quantum interval
...
+    # parameters are optional, and can be specified as empty or nil strings.
+    # @private
+    def set_tls_parms signature, priv_key, cert_chain, verify_peer, fail_if_no_peer_cert, sni_hostname, cipher_list, ecdh_curve, dhparam, protocols_bitmask
+      bitmask = protocols_bitmask
+      ssl_options = OpenSSL::SSL::OP_ALL
+      ssl_options |= OpenSSL::SSL::OP_NO_SSLv2 if defined?(OpenSSL::SSL::OP_NO_SSLv2) && EM_PROTO_SSLv2 & bitmask == 0
+      ssl_options |= OpenSSL::SSL::OP_NO_SSLv3 if defined?(OpenSSL::SSL::OP_NO_SSLv3) && EM_PROTO_SSLv3 & bitmask == 0
+      ssl_options |= OpenSSL::SSL::OP_NO_TLSv1 if defined?(OpenSSL::SSL::OP_NO_TLSv1) && EM_PROTO_TLSv1 & bitmask == 0
+      ssl_options |= OpenSSL::SSL::OP_NO_TLSv1_1 if defined?(OpenSSL::SSL::OP_NO_TLSv1_1) && EM_PROTO_TLSv1_1 & bitmask == 0
+      ssl_options |= OpenSSL::SSL::OP_NO_TLSv1_2 if defined?(OpenSSL::SSL::OP_NO_TLSv1_2) && EM_PROTO_TLSv1_2 & bitmask == 0
```
portertech: @sodabrew I applied your patch.

lib/em/pure_ruby.rb
sodabrew: I think this is OK? I wonder if it will work-alike.

lib/em/pure_ruby.rb
sodabrew: Thanks!

lib/em/pure_ruby.rb
sodabrew: The problem is that this will rescue instances of OpenSSL::SSL::SSLError in the blocks for the Read/Write errors. I think the way to roll is to flip the situation around, so below you'd rescue a new class `EventMachine::Error::SSLErrorWaitWritable`, and then define it as:

``` ruby
if defined?(OpenSSL::SSL::SSLErrorWaitWritable)
  EventMachine::Error::SSLErrorWaitWritable = OpenSSL::SSL::SSLErrorWaitWritable
else
  class EventMachine::Error::SSLErrorWaitWritable < EventMachine::Error; end
end
```

portertech: This is a better approach, I'll apply a similar patch :+1:

portertech: Figuring out what openssl threw before wait writable etc.

lib/em/pure_ruby.rb
```diff
@@ -33,6 +33,36 @@
...
+  else
+    SSLConnectionWaitReadable = IO::WaitReadable
+  end
+
+  # Older versions of Ruby may not provide the SSLErrorWaitWritable
+  # OpenSSL class. Create an error class to act as a "proxy".
+  if defined?(OpenSSL::SSL::SSLErrorWaitWritable)
+    SSLConnectionWaitWritable = OpenSSL::SSL::SSLErrorWaitWritable
+  else
+    SSLConnectionWaitWritable = IO::WaitWritable
```
portertech: I have tested on Ruby 2.0.0 and 2.3.0 :+1:

portertech:
@sodabrew ready for another go :+1:

lib/em/pure_ruby.rb
portertech: Actually, it works with a client, but the specs do not work.

lib/em/pure_ruby.rb
```diff
@@ -33,6 +33,88 @@
...
+      @cert.add_extension factory.create_extension("authorityKeyIdentifier", "keyid:always,issuer:always")
+      @cert.sign(@key, OpenSSL::Digest::SHA1.new)
+    end
+  end
+
+  # @private
+  DefaultCertificate = CertificateCreator.new
+
+  # @private
+  DefaultDHKey1024 = OpenSSL::PKey::DH.new <<-_end_of_pem_
```
portertech: The default DH key params are from Ruby 2.3.0 stdlib.

portertech:
@sodabrew ready for your review & testing once again :+1:

lib/em/pure_ruby.rb
sodabrew: Super awesome, thanks for chasing down the older exception!

portertech:
@sodabrew please let me know if there are any other changes that I need to make. We would love to see this make it into upstream :+1:

sodabrew:
Would you mind squashing to fewer commits? I think just a few commits should suffice for the final form of this PR.

portertech:
@sodabrew commits clean up :+1:

sodabrew:
Thank you!

Issue #633, opened 2015-08-29
[1.0.8] kqueue regression:  arm kqueue writer failed


felixbuenemann:
I am experiencing crashes when using thin 1.6.3 with kqueue on Mac OS X 10.11 (15A263e) and ruby 2.2.3p173 x86_64 which seemed to be cause by a change introduced in eventmachine 1.0.8, because they do not occur on 1.0.7.

Crash using a [testcase with an async thin app](https://gist.github.com/ea29903e18e328cd9e8b):
```
./em-108-arm-kqueue-writer-failed-testcase.rb
Using eventmachine 1.0.8
Using rack 1.6.4
Using bundler 1.10.6
Using daemons 1.2.3
Using thin 1.6.3
Thin web server (v1.6.3 codename Protein Powder)
Maximum connections set to 1024
Listening on 127.0.0.1:3000, CTRL+C to stop
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: arm kqueue writer failed on 32681: Bad file descriptor
[1]    67463 abort      ./em-108-arm-kqueue-writer-failed-testcase.rb
```
The app was hit using `ab -r -c 150 -n 15000 http://127.0.0.1:3000/`, for details see the [gist comment](https://gist.github.com/felixbuenemann/ea29903e18e328cd9e8b#gistcomment-1562436).

I've extracted a backtrace of the crash using gdb:
```
#0  0x0000000000000100 in ?? ()
#1  0x0000000102a7e2ac in EventMachine_t::_RunKqueueOnce() ()
   from /Users/felix/.rvm/gems/ruby-2.2.3/gems/eventmachine-1.0.8/lib/rubyeventmachine.bundle
#2  0x0000000102a7d3f0 in EventMachine_t::RunOnce() ()
   from /Users/felix/.rvm/gems/ruby-2.2.3/gems/eventmachine-1.0.8/lib/rubyeventmachine.bundle
#3  0x0000000102a7d33d in EventMachine_t::Run() ()
   from /Users/felix/.rvm/gems/ruby-2.2.3/gems/eventmachine-1.0.8/lib/rubyeventmachine.bundle
#4  0x0000000102a69c6c in evma_run_machine ()
   from /Users/felix/.rvm/gems/ruby-2.2.3/gems/eventmachine-1.0.8/lib/rubyeventmachine.bundle
#5  0x0000000102aa29c1 in t_run_machine(unsigned long) ()
[...]

PR #728, opened 2016-08-08
Fix deregistering descriptor when using KQUEUE

luismfonseca:
It was possible that a write was planned to happen but the socket got closed
in the meantime. This would lead to a serious error as it would then attempt
to write on closed socket (-1 socket) and crash.

To be more specific, the error I got at the application was:
```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: arm kqueue writer failed on -1: Bad file descriptor
```

It's not clear to me where and how to write a spec for this (this only manifests on a KQUEUE system), please let me know how to proceed.

### DIFF
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -1889,6 +1889,14 @@ void EventMachine_t::Deregister (EventableDescriptor *ed)
 		ModifiedDescriptors.erase(ed);
 	}
 	#endif
+
+	#ifdef HAVE_KQUEUE
+	if (Poller == Poller_Kqueue) {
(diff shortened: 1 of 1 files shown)


sodabrew:
Sorry for the delay getting back to you. I may have seen this crash before on OS X. Let me think about about to make a test for it.

luismfonseca:
Thank you.

For reference: the fix I did solve my problem. It looks more of an oversight because I don't think there's a reason for the (essentially the same) logic not be implemented for KQUEUE.

sodabrew:
It turns out that this resolves #633!

PR #731, opened 2016-08-29
Fix to enable to work an example code in EM::Pool

taise:
I fixed an example code  in EventMachine::Pool
- It was expected to request `100.times`, but it was only one requested
- Exception: `undefined local variable or method i`

### DIFF
--- a/lib/em/pool.rb
+++ b/lib/em/pool.rb
@@ -6,6 +6,8 @@ module EventMachine
   #
   # Example:
   #
+  #    require 'em-http-request'
+  #
   #    EM.run do
   #      pool  = EM::Pool.new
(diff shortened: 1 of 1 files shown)


lib/em/pool.rb
```diff
@@ -1,41 +1,41 @@
 module EventMachine
-  # = EventMachine::Pool
```
sodabrew: Might need to keep this line for the documentation to generate?

taise: Thank you for checking.
It seems style of Rdoc.
When I executed `bundle exec rake yard`, `doc/EventMachine/Pool.html` was generated.

sodabrew: Ok! Thanks for checking. Thanks for updating the example!

PR #732, opened 2016-09-06
Fix deregistering descriptor when using KQUEUE

luismfonseca:
It was possible that a write was planned to happen but the socket got closed
in the meantime. This would like to a serious error as it would then attempt
to write on closed socket (-1 socket) and crash.


luismfonseca:
Sorry, duplicated PR.

PR #736, opened 2016-10-04
Add ability to subscribe to multicast channels

poulh:
With this change I can subscribe to multicast data by mcastjoining the passed in host/port.

before i could not access the socket to do the mcast subscription.

### DIFF
--- a/lib/em/version.rb
+++ b/lib/em/version.rb
@@ -1,3 +1,3 @@
 module EventMachine
-  VERSION = "1.2.0.1"
+  VERSION = "1.2.0.2"
 end
--- a/lib/eventmachine.rb
+++ b/lib/eventmachine.rb
@@ -881,6 +881,19 @@ module EventMachine
     c
   end
 
+  # Same as open_datagram_socket but also multicast subscribes
+  def self.open_datagram_socket_and_mcastjoin address, port, handler=nil, *args
+
+    klass = klass_from_handler(Connection, handler, *args)
+    s = EM::open_udp_socket address, port.to_i
+    ip =  IPAddr.new( address ).hton + IPAddr.new("0.0.0.0").hton
+    set_sock_opt s, Socket::IPPROTO_IP, Socket::IP_ADD_MEMBERSHIP, ip
+    c = klass.new s, *args
+    @conns[s] = c
+    block_given? and yield c
+    c
+  end
+
 
   # For advanced users. This function sets the default timer granularity, which by default is
   # slightly smaller than 100 milliseconds. Call this function to set a higher or lower granularity.


sodabrew:
Thanks! Reviewing today!

lib/em/version.rb
```diff
@@ -1,3 +1,3 @@
 module EventMachine
-  VERSION = "1.2.0.1"
+  VERSION = "1.2.0.2"
```
sodabrew: Please don't bump the version in a PR - I'll do that when I release a group of changes.

lib/eventmachine.rb
```diff
@@ -881,6 +881,19 @@ def self.open_datagram_socket address, port, handler=nil, *args
     c
   end
 
+  # Same as open_datagram_socket but also multicast subscribes
+  def self.open_datagram_socket_and_mcastjoin address, port, handler=nil, *args
+
+    klass = klass_from_handler(Connection, handler, *args)
+    s = EM::open_udp_socket address, port.to_i
```
sodabrew: You don't need the `EM::` namespace since you're already inside `EM`

poulh:
Hi,

It turns out you can do this without my commit.  I didn't realize it until after I sent the pull request.  The following code works with the existing open_datagram_socket method
```

 udp_handler = EM::open_datagram_socket(host, port, MulticastHandler)
 ip =  IPAddr.new( host ).hton + IPAddr.new("0.0.0.0").hton
 udp_handler.set_sock_opt Socket::IPPROTO_IP, Socket::IP_ADD_MEMBERSHIP, ip
```

Maybe its worth putting this in an example?  I found it difficult to find the answer googling around.

lib/eventmachine.rb
```diff
@@ -881,6 +881,19 @@ def self.open_datagram_socket address, port, handler=nil, *args
...
+    ip =  IPAddr.new( address ).hton + IPAddr.new("0.0.0.0").hton
```
sodabrew: I don't remember enough about multicast if the interface address should be configurable here or just use the any address?

lib/eventmachine.rb
```diff
@@ -881,6 +881,19 @@ def self.open_datagram_socket address, port, handler=nil, *args
...
+    set_sock_opt s, Socket::IPPROTO_IP, Socket::IP_ADD_MEMBERSHIP, ip
```
sodabrew: I'm a little worried about this creating a precedent for adding separate top-level methods for each sockopt you might want to set. Could you do the IP_ADD_MEMBERSHIP in the post_init of your connection handler class?

sodabrew: Poke @poulh

sodabrew:
Per comment https://github.com/eventmachine/eventmachine/pull/736#issuecomment-251477301 this PR isn't required because you can call set_sock_opt on the underlying socket with arguments to join a multicast channel.

PR #741, opened 2016-11-04
Pure Ruby EM, rescue ECONNREFUSED on initial TCP connect

portertech:


sodabrew:
In the case of the ECONNREFUSED, it'll continue to pass the unconnected socket to the EvmaTCPClient class. Is that better than throwing the exception in the connect method?

PR #471, opened 2013-10-11
Add TLS support to Java reactor

MattVonVielen:
This branch adds SSL/TLS support to the Java reactor.  All ssl-related tests pass against JRuby (though there are still many unrelated test failures on JRuby which exist on master).

Addresses issues #34, #180, #379, and one line-item of #117

PR #743, opened 2016-11-07
Use Acceptors to get peer and sock names if not present in Connections

lampad:
On Jruby, when starting a tcp server locally using the 0 port argument (to select a random port), I was later trying to discover what port it was on by calling: `Socket.unpack_sockaddr_in( EM.get_sockname( signature ) ).first`

Because the server doesn't get added to the Connections HashMap, I was getting an NPE. This PR appears to address it in my local tests, and it favors entries in Connections and falls back to the Acceptors.

I also added a couple things to make java builds friendlier, and a missing constant that was also causing me some pains.

Things appear to install and run fine, but when I try to run tests (using JRuby 9.1.2.0), I end up getting many errors about too many files being open. I could crank up my ulimit (it's already at 512), but I'm not sure how much further I need to go.

### DIFF
--- a/eventmachine.gemspec
+++ b/eventmachine.gemspec
@@ -12,12 +12,7 @@ Gem::Specification.new do |s|
   s.files = `git ls-files README.md CHANGELOG.md GNU LICENSE rakelib ext java lib docs`.split
   s.test_files = `git ls-files tests examples`.split
 
-  if RUBY_PLATFORM =~ /java/
-    s.files << "lib/rubyeventmachine.jar"
-    s.platform = 'java'
-  else
-    s.extensions = ["ext/extconf.rb", "ext/fastfilereader/extconf.rb"]
-  end
+  s.extensions = ["ext/extconf.rb", "ext/fastfilereader/extconf.rb"]
 
   s.add_development_dependency 'test-unit', '~> 2.0'
   s.add_development_dependency 'rake-compiler', '~> 0.9.5'
--- a/eventmachine.gemspec
+++ b/eventmachine.gemspec
@@ -12,7 +12,12 @@ Gem::Specification.new do |s|
   s.files = `git ls-files README.md CHANGELOG.md GNU LICENSE rakelib ext java lib docs`.split
   s.test_files = `git ls-files tests examples`.split
 
-  s.extensions = ["ext/extconf.rb", "ext/fastfilereader/extconf.rb"]
+  if RUBY_PLATFORM =~ /java/
+    s.files << "lib/rubyeventmachine.jar"
+    s.platform = 'java'
+  else
+    s.extensions = ["ext/extconf.rb", "ext/fastfilereader/extconf.rb"]
+  end
 
   s.add_development_dependency 'test-unit', '~> 2.0'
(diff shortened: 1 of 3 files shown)


eventmachine.gemspec
```diff
@@ -12,7 +12,12 @@ Gem::Specification.new do |s|
   s.files = `git ls-files README.md CHANGELOG.md GNU LICENSE rakelib ext java lib docs`.split
   s.test_files = `git ls-files tests examples`.split
 
-  s.extensions = ["ext/extconf.rb", "ext/fastfilereader/extconf.rb"]
+  if RUBY_PLATFORM =~ /java/
+    s.files << "lib/rubyeventmachine.jar"
+    s.platform = 'java'
+  else
+    s.extensions = ["ext/extconf.rb", "ext/fastfilereader/extconf.rb"]
+  end
```
sodabrew: I don't recall how conditionals in the gemspec interact with rubygems.org?

lampad: Hmm... A good question. I'm not sure. I wasn't aware that there were any interactions between specs and rubygems.org. I thought it was all evaluated at ruby runtime. A quick google doesn't seem to bring up anything obvious.

sodabrew: What prompted your change here? I'm not a regular JRuby user, but let me know what the problem was / improvement is for?

lampad: Setting the platform and adding the jar file will build a 'java' gem that will be preferentially searched for installed if you're on a java platform when it's available on rubygems.

Segregating out the extensions will prevent jruby from trying to build those extensions (which it can't do) so that the gem can be installed.

lampad:
I've got a couple more changes coming to support TLS connections (a few more missing constants, and setting some tls opts.

eventmachine.gemspec
sodabrew: I've built and uploaded the Java gem without this, are you sure it's necessary?

sodabrew: Thanks @tmm1! So @lampad Would you pull this commit out from the PR and then I'll merge the Acceptor changes.

lampad: I had been using `rake gem` and `rake package` and they didn't seem to be java-specific (it was still trying to build c-extensions when I installed the local one), but I didn't try `rake build`. Also, entirely possible I was doing something dumb.

I'd be happy to pull that out. I found a few more things I'd like to contribute. Would you prefer those in a separate PR, or in this same one?

tmm1: Sorry `rake gem` is correct, but you would need to ensure you were using jruby+rake to invoke it.

sodabrew: @lampad Before I build the 1.2.1 gem, I'll be sure to revisit this to make sure I can build the JRuby gem.

lampad:
The other TLS changes I mentioned will come in another pull request, as there's some work to be done on the java side to support it.

sodabrew:
@lampad Speaking of Java TLS, take a look at #471 and see if some of that can be forward-ported to the current master branch?

lampad:
@sodabrew Sure. Looks good at first glance. I'll dig a bit deeper when I get some free time.

PR #755, opened 2016-12-27
Update extconf.rb

krzcho:
updated library filenames to compile with fresh OpenSSL on Windows

### DIFF
--- a/ext/extconf.rb
+++ b/ext/extconf.rb
@@ -27,7 +27,7 @@ end
 
 SSL_HEADS = %w(openssl/ssl.h openssl/err.h)
 SSL_LIBS = case RUBY_PLATFORM
-when /mswin|mingw|bccwin/ ; %w(ssleay32 libeay32)
+when /mswin|mingw|bccwin/ ; %w(libcrypto libssl)
 else                      ; %w(crypto ssl)
 end
 


sodabrew:
Thanks for the suggested change! I tried, this but it doesn't work for me with the OpenSSL 1.1.0c Light distribution. The library files I see are named `libcrypto-1_1-x64.dll` and `libssl-1_1-x64.dll`. Which OpenSSL binary distribution did you install? What are the files in its main directory?

I don't want to drop compatibility with the 1.0.2 DLLs, either, so the change will need to be a little more involved than this initial PR.

krzcho:
"Win32 OpenSSL v1.1.0c	30MB Installer	Installs Win32 OpenSSL v1.1.0c (Recommended for software developers by the creators of OpenSSL)"
I got libcrypto-1_1.a, libcrypto-1_1.def, libssl-1_1.a and libssl-1_1.def in lib\MinGW - I guess those are (statically) linked, not dlls you mentioned

sodabrew:
The DLLs are in the top-level of the OpenSSL distribution directory.

sodabrew:
Hmm, static linking is an idea. I did similar work with the .def files for the MySQL Connector/C in the mysql2 gem. That would basically fix this problem by distributing EventMachine for Windows with OpenSSL baked in.

krzcho:
I have checked using system profiler (procmon.exe) that linker (ld.exe) is actually using
C:\OpenSSL\lib\libcrypto.lib and C:\OpenSSL\lib\libssl.lib
It is only searching for libs in \lib subdirectory of OpenSSL
It is trying to use dll first by searching for liblibssl.dll.a but that is not available in OpenSSL distibution
I know the difference between static and dynamic linking but not much more to help build EM with OpenSSL dlls but for me static linking should be more than fine

MSP-Greg:
@krzcho I apologize for not seeing this PR.  I submitted [785](https://github.com/eventmachine/eventmachine/pull/785) for this same issue, but tested using a MinGW package (which is the same package I build Ruby OpenSSL with.)

If you have a minute, I'd appreciate if you tried the patch I posted, as I believe it should work with most versions of OpenSSL...

krzcho:
No problem; you are still using (ssleay32 libeay32) while I have changed that into (libcrypto libssl) from latest openssl

MSP-Greg:
@krzcho 

The patch in [785](https://github.com/eventmachine/eventmachine/pull/785) should allow both new (1.1.0) and old (1.02 and earlier) to be used.  I used the patch with 1.1.0f.

sodabrew:
Closing in favor of #785

Issue #512, opened 2014-05-16
EM crashes on Linux systems when calling stop_watching if there have been multiple .watch_file calls to the same file


rb2k:
This works (given that /tmp/foo exists):

``` ruby
require 'rubygems'
require 'eventmachine'
EM.kqueue=true
EventMachine.run {
  w1 = EventMachine.watch_file("/tmp/foo")
  w2 = EventMachine.watch_file("/tmp/foo")
  w2.stop_watching
}
```

This fails:

``` ruby
require 'rubygems'
require 'eventmachine'
EventMachine.run {
  w1 = EventMachine.watch_file("/tmp/foo")
  w2 = EventMachine.watch_file("/tmp/foo")
  w2.stop_watching
}


#  ruby emtest.rb
terminate called after throwing an instance of 'std::runtime_error'
  what():  attempted to remove invalid watch signature
Aborted
```

Notice that the return value from inotify_add_watch is the same (1):

```
stat("/tmp/foo", {st_mode=S_IFREG|0644, st_size=0, ...}) = 0
inotify_init()                          = 6
fcntl(6, F_GETFL)                       = 0 (flags O_RDONLY)
fcntl(6, F_SETFL, O_RDONLY|O_NONBLOCK)  = 0
inotify_add_watch(6, "/tmp/foo", IN_MODIFY|IN_MOVED_FROM|IN_MOVED_TO|IN_CREATE|IN_DELETE|IN_DELETE_SELF|IN_MOVE_SELF) = 1
stat("/tmp/foo", {st_mode=S_IFREG|0644, st_size=0, ...}) = 0
inotify_add_watch(6, "/tmp/foo", IN_MODIFY|IN_MOVED_FROM|IN_MOVED_TO|IN_CREATE|IN_DELETE|IN_DELETE_SELF|IN_MOVE_SELF) = 1
```

EM uses a c++ std::map to map the inotify watch descriptor (wd) to the "bindable" object.
std::map has unique keys and Eventmachine [uses std::map::insert()](https://github.com/eventmachine/eventmachine/blob/2ed44025e35518898c8589a610515e54a7b40c44/ext/em.cpp#L2176-L2177)

Looking for this map insert function leads to:
"Because element keys in a map are unique, the insertion operation checks whether each inserted element has a key equivalent to the one of an element already in the container, and if so, the element is not inserted."

This is why unwatching the _second_ open watcher always fails.
[...]

PR #757, opened 2016-12-29
Raise a ruby exception for InvalidWatchSignature

surenm:
Addresses #512

Some how the runtime error thrown in evma_unwatch_filename was not caught and EM crashed with std::runtime_error. A Simple patch to fix this by throwing a ruby exception instead.

Not sure if this is the correct fix but at least the ruby exception can be caught and handled further.

### DIFF
--- a/ext/rubymain.cpp
+++ b/ext/rubymain.cpp
@@ -58,7 +58,7 @@ static VALUE EM_eConnectionError;
 static VALUE EM_eUnknownTimerFired;
 static VALUE EM_eConnectionNotBound;
 static VALUE EM_eUnsupported;
-static VALUE EM_eInvalidWatchSignature;
+static VALUE EM_eInvalidSignature;
 
 static VALUE Intern_at_signature;
 static VALUE Intern_at_timers;
@@ -1040,7 +1040,7 @@ static VALUE t_unwatch_filename (VALUE self UNUSED, VALUE sig)
 	try {
 		evma_unwatch_filename(NUM2BSIG (sig));
 	} catch (std::runtime_error e) {
-		rb_raise (EM_eInvalidWatchSignature, "%s", e.what());
+		rb_raise (EM_eInvalidSignature, "%s", e.what());
 	}
 	
 	return Qnil;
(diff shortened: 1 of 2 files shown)


surenm:
Not sure why the travis build failed either, didn't seem like a related failure. If it is related, can you please help me with fixing it?

sodabrew:
The build failure was unrelated, I just re-ran the build so you can see your code turns up green.

This looks good! I have a suggestion/request for a change: The signatures system is used by more than just file watching, so I think the exception should be called `InvalidSignature` and then it can use it everywhere that a signature argument is used.

surenm:
Thanks for getting back quickly! Made the refactor and added a unit test case as well 👍

tests/test_file_watch.rb
```diff
@@ -55,6 +55,25 @@ def test_events
...
+    # Refer: https://github.com/eventmachine/eventmachine/issues/512
+    def test_invalid_signature
+      # This works fine with kqueue, only fails with linux inotify.
+      omit_if(EM.kqueue?)
+
+      EM.run {
+        file = Tempfile.new('foo')
+
+        w1 = EventMachine.watch_file(file.path)
+        w2 = EventMachine.watch_file(file.path)
```
sodabrew: This returns a real Watcher object that isn't valid? That seems bad, like, the better solution here is to return `nil` if the file is already being watched and we can't watch it a second time.

In the case of kqueue, does it actually create two independent watchers?

ext/rubymain.cpp
```diff
@@ -1036,7 +1037,12 @@ t_unwatch_filename
 
 static VALUE t_unwatch_filename (VALUE self UNUSED, VALUE sig)
 {
-	evma_unwatch_filename(NUM2BSIG (sig));
+	try {
+		evma_unwatch_filename(NUM2BSIG (sig));
+	} catch (std::runtime_error e) {
+		rb_raise (EM_eInvalidSignature, "%s", e.what());
+	}
```
sodabrew: I agree with this regardless of the reason for the invalid signature; raising a C++ exception into Ruby space that kills the application isn't helpful to anybody.

tests/test_file_watch.rb
surenm: The watcher it returns is valid but the binding to watch map that is used to keep reference of the watch is not properly updated, so stopping the watch raises this error.

The inotify and kqueue behaviour is reported more thoroughly in #512.

PR #762, opened 2017-01-29
EM::Connection#watch_only?

genki:
This PR provides a way to test if a connection is watch_only.

### DIFF
--- a/ext/cmain.cpp
+++ b/ext/cmain.cpp
@@ -360,6 +360,19 @@ extern "C" void evma_unwatch_pid (const uintptr_t sig)
 	EventMachine->UnwatchPid(sig);
 }
 
+/*****************
+evma_is_watch_only
+*****************/
+
+extern "C" int evma_is_watch_only (const uintptr_t binding)
+{
+	EventableDescriptor *cd = dynamic_cast <EventableDescriptor*> (Bindable_t::GetObject (binding));
+	if (cd)
+    return cd->IsWatchOnly() ? 1 : 0;
+
+	return 0;
+}
+
 /****************************
 evma_send_data_to_connection
 ****************************/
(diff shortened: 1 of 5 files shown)


ext/cmain.cpp
```diff
@@ -360,6 +360,19 @@ extern "C" void evma_unwatch_pid (const uintptr_t sig)
...
 
+/*****************
+evma_is_watch_only
+*****************/
+
+extern "C" int evma_is_watch_only (const uintptr_t binding)
+{
+	EventableDescriptor *cd = dynamic_cast <EventableDescriptor*> (Bindable_t::GetObject (binding));
+	if (cd)
+    return cd->IsWatchOnly() ? 1 : 0;
```
sodabrew: This looks like a spaces vs. tabs mismatch.

genki:
My use case is to list up all EM connections and close them safely while a sort of finalization process.
Current implementation throws C-level exception (hence causes a crash) if the connection to be closed is the watchonly, even there is no way to check if it is watchonly.
By using this PR, we can check it before calling close to avoid a panic.

ext/cmain.cpp
genki: Changed spaces to TAB.

sodabrew:
On what platform does it crash when closing the (file-?)watch? If it's Linux only then this is actually a known issue that is fixed in 1.2.2 and I'll upload it in just a minute.

sodabrew:
You might be experiencing issue #512. Change #757 in the just-released EventMachine 1.2.2 catches the C++ exception and raises a rescueable Ruby exception. Please give it a try!

genki:
@olegshaldybin Thanks, I will try.

sodabrew:
Let me know if EventMachine 1.2.2 helped your situation @genki? If this new API would still be helpful to your program, I want to make sure to add the API to docs and other reactors as well.

genki:
@sodabrew Sorry for delayed response.
Your fix looks great but  there is no chance to know which method should be called either `close` or `stop_watching`, for the listed connection objects without using `watch_only?`

Issue #765, opened 2017-02-01
Segfault from EM.run after EM::Connection#unbind raises exception


SpComb:
I experienced a segfault similar to  #758 and  #761 when trying to write rspecs for a `faye/websocket` client, using an `EM.run { ... }` pattern similar to [`tests/test_unbind_reason.rb`](https://github.com/eventmachine/eventmachine/blob/master/tests/test_unbind_reason.rb).

The following minimal repro crashes with a segfault when run with `CRASH_EM=1`:

```ruby
#!/usr/bin/env ruby

require 'eventmachine'

class Subject < EM::Connection
  def self.connect(host, port)
    STDERR.puts "EM.connect..."

    EM.connect '127.0.0.1', 1337, self
  end

  def unbind(reason = nil)
    STDERR.puts "EM::Connection.unbind: #{reason}"

    fail 'crash EventMachine' if ENV['CRASH_EM']

    EM.stop
  rescue => exc
    STDERR.puts "raise #{exc}:\n#{exc.backtrace.join "\n"}\n\n"
    raise
  end
end

STDERR.puts "EM.run..."
EM.run {
  STDERR.puts "EM.run &block start"
  conn = Subject.connect('127.0.0.1', 1337)
  STDERR.puts "EM.run &block end"
}
STDERR.puts "EM.run returned"
```

#### `ruby bin/em-test`
```
EM.run...
EM.run &block start
EM.connect...
EM.run &block end
EM::Connection.unbind: Errno::ECONNREFUSED
EM.run returned
```

#### `CRASH_EM=1 ruby bin/em-test`
```
EM.run...
EM.run &block start
EM.connect...
EM.run &block end
EM::Connection.unbind: Errno::ECONNREFUSED
raise crash EventMachine:
bin/em-test:15:in `unbind'
/home/kontena/.gem/ruby/2.3.0/gems/eventmachine-1.2.2/lib/eventmachine.rb:1481:in `event_callback'
/home/kontena/.gem/ruby/2.3.0/gems/eventmachine-1.2.2/lib/eventmachine.rb:194:in `run_machine'
/home/kontena/.gem/ruby/2.3.0/gems/eventmachine-1.2.2/lib/eventmachine.rb:194:in `run'
bin/em-test:25:in `<main>'

/home/kontena/.gem/ruby/2.3.0/gems/eventmachine-1.2.2/lib/eventmachine.rb:202: [BUG] Segmentation fault at 0x00000000000000
[...]

Issue #758, opened 2017-01-17
Segmentation Fault eventmachine-1.2.1/lib/eventmachine.rb:202


Bogadon:
This is probably a duplicate of similar issues, but uploading this trace in case its useful.

We have been seeing this error very frequently. Unfortunately I can not speculate on the circumstance; sometimes it lasts for days, other times it crashes on start up several times in a row, then lasts for several hours.

If you need any other information I can try to provide.

Thanks for the gem.

> Thin web server (v1.7.0 codename Dunder Mifflin)
> Maximum connections set to 1024
> Listening on 0.0.0.0:3001, CTRL+C to stop
> /home/yozu/shared/bundle/ruby/2.3.0/gems/eventmachine-1.2.1/lib/eventmachine.rb:202: [BUG] Segmentation fault at 0x00000000000000
> ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-linux]
> 
> -- Control frame information -----------------------------------------------
> c:0046 p:---- s:0192 e:000191 CFUNC  :release_machine
> c:0045 p:0036 s:0189 e:000188 RESCUE /home/yozu/shared/bundle/ruby/2.3.0/gems/eventmachine-1.2.1/lib/eventmachine.rb:202
> c:0044 p:0385 s:0186 E:0010b0 METHOD /home/yozu/shared/bundle/ruby/2.3.0/gems/eventmachine-1.2.1/lib/eventmachine.rb:205
> c:0043 p:0063 s:0179 E:000868 METHOD /home/yozu/shared/bundle/ruby/2.3.0/gems/thin-1.7.0/lib/thin/backends/base.rb:73
> c:0042 p:0120 s:0175 E:000d10 METHOD /home/yozu/shared/bundle/ruby/2.3.0/gems/thin-1.7.0/lib/thin/server.rb:162
> c:0041 p:0511 s:0172 E:001798 METHOD /home/yozu/shared/bundle/ruby/2.3.0/gems/thin-1.7.0/lib/thin/controllers/controller.rb:87
> c:0040 p:0023 s:0168 e:000167 BLOCK  /home/yozu/shared/bundle/ruby/2.3.0/bundler/gems/websocket-rails-08b3b3b13b7b/lib/rails/tasks/websocket_rails.tasks:14 [FINISH]

Full log attached
[eventmachinefailure.txt](https://github.com/eventmachine/eventmachine/files/711306/eventmachinefailure.txt)

[...]

PR #766, opened 2017-02-01
Fix segfault when an Exception is raised from unbind callback

arthurdandrea:
I experienced a segfault similar to #765, #758 and #761, debugging the code I found that the problem was caused by a double free of the a `EventableDescriptor`.

In detail `EventMachine_t::_CleanupSockets()` was calling the `EventableDescriptor` destructor that triggers the `Connection#unbind` callback. The problem is that when the callback raises an exception, it stops the EventMachine and eventually calls the `EventMachine_t` destructor that attempts to destroy the same `EventableDescriptor` since the `EventMachine_t::_CleanupSockets()` function hasn't yet cleaned up the Descriptors vector.
The workaround used was very crude but it works, it delays the destruction of the `EventableDescriptor` to after they are removed from the `Descriptors` vector.

### DIFF
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -164,6 +164,8 @@ EventMachine_t::~EventMachine_t()
 {
 	// Run down descriptors
 	size_t i;
+	for (i = 0; i < DescriptorsToDelete.size(); i++)
+ 		delete DescriptorsToDelete[i];
 	for (i = 0; i < NewDescriptors.size(); i++)
 		delete NewDescriptors[i];
 	for (i = 0; i < Descriptors.size(); i++)
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -840,6 +840,17 @@ void EventMachine_t::_CleanupSockets()
 		EventableDescriptor *ed = Descriptors[i];
 		assert (ed);
 		if (ed->ShouldDelete()) {
+			DescriptorsToDelete.push_back(ed);
+		}
+		else
+			Descriptors [j++] = ed;
+	}
+	while ((size_t)j < Descriptors.size())
+		Descriptors.pop_back();
+
+	nSockets = DescriptorsToDelete.size();
+	for (i=0; i < nSockets; i++) {
+		EventableDescriptor *ed = DescriptorsToDelete[i];
 		#ifdef HAVE_EPOLL
 			if (Poller == Poller_Epoll) {
 				assert (epfd != -1);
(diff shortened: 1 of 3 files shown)


sodabrew:
Thanks for the contribution! This will certainly solve the problem of freeing the same descriptor twice, and in fact that might be the root of a number of open issues as you linked. But if one of them raises an exception it would seem that the others will be leaked?

I wonder if the `delete ed` line can be wrapped to catch a Ruby-level exception.

sodabrew:
I traced the path of the destructor, through the `event_callback` function in rubymain.cpp, which does call unbind through an `rb_rescue()`. What's not clear to be is if we ever re-enter the destructor in C++ land to finish the process of unbinding all of the descriptors. If we don't that actually means we can trigger a leak this way, and the rb_rescue also needs an rb_ensure wrapper to make sure we go back and finish the cleanup job.

arthurdandrea:
I think the best solution would be not to call any Ruby code from destructors, but apparently it would require more changes on how the `unbind`, `proxy_target_unbound` and possibly others callbacks are fired. Regardless of the ideal solution, I agree that extra care should be taken around the C++ <-> Ruby crossings to avoid any memory leak or segfaults in corner cases such as this one.

sodabrew:
I tried two different alternative approaches in #768, but I think your approach here remains the best option.

On review, I think you're missing one thing, which is in the `EventMachine::~EventMachine` destructor, there should be a loop like this:

```
EventMachine_t::~EventMachine_t()
{
...
 	for (i = 0; i < DescriptorsToDelete.size(); i++)
 		delete DescriptorsToDelete[i];
...
}
```

sodabrew:
An area I still need to get clear on, at the point of the EventMachine destructor, if an unbind _still_ throws an exception, we need to jump back into the destructor so that all memory is freed. The `rb_protect` function is designed for this, but there are unit tests which have the expectation of being able to catch an exception in lexical scope of the EM.run { ... } that ends up outside of the dynamic scope of the event loop.

Leaking memory might be less-bad than outright crashing, but because `EM.run { ... }` loops can be executed more than once in a ruby process, if each invocation of EM.run can leak memory the process can end up in a bad spot.

Setting an `EM.error_handler { ... }` block may mitigate this problem.

sodabrew:
(I might take one more stab at only preventing exceptions from `unbind` from bubbling all the way up, because that callback is specifically important for resource cleanup from C++ space.)

arthurdandrea:
Inspired by your work on #768 I'm creating a new pull request where all callbacks are wrapped in rb_rescue (I think its coming to be a better solution even than this pull request), meanwhile in the next hours i'll update this pull request with the missing code from `EventMachine_t::~EventMachine_t`.

arthurdandrea:
Ooops, wrong button!

sodabrew:
Thank you!

PR #767, opened 2017-02-08
Allow destructors to throw when compiling in >= C++11

Peter-Levine:
In C++11 and later, destructors are marked `noexcept(true)` by default.  All throws become calls to `terminate()`.  To enable the original intent of the throw, such destructors (and if virtual, those of its parent classes) need to be explicitly marked `noexcept(false)` when compiled in a dialect >= C++11.

### DIFF
--- a/ext/binder.cpp
+++ b/ext/binder.cpp
@@ -116,10 +116,7 @@ Bindable_t::Bindable_t()
 Bindable_t::~Bindable_t
 ***********************/
 
-Bindable_t::~Bindable_t()
-#if __cplusplus >= 201103L
-noexcept(false)
-#endif
+Bindable_t::~Bindable_t() NO_EXCEPT_FALSE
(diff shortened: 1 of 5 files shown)


sodabrew:
Thanks! Interesting change in C++11. Is it possible to macro this in one place, for example:

In a header file:
```
#if __cplusplus >= 201103L
#define NO_EXCEPT_FALSE noexcept(false)
#else
#define NO_EXCEPT_FALSE
#endif
```

Everywhere else:
```
Bindable_t::~Bindable_t()
NO_EXCEPT_FALSE
  {
  	BindingBag.erase (Binding);
  }
```

Peter-Levine:
> Is it possible to macro this in one place...

Yeah.  Updated it.  Putting it  `ext/binder.h` works fine on my end.

PR #768, opened 2017-02-09
Wrap all callbacks with rb_protect to prevent raising exceptions

sodabrew:
This might work for #766


sodabrew:
Ok this breaks test_bubble_errors_from_initialize.

Maybe there needs to be a flag indicating if an exception can be propagated or not, i.e. for the Unbind case.

sodabrew:
Closed in favor of #766

PR #771, opened 2017-02-17
Wrap all callbacks with rb_rescue to prevent Ruby exceptions to disturb C++ land

arthurdandrea:
Alternative to #766 and #768.

### DIFF
--- a/lib/em/connection.rb
+++ b/lib/em/connection.rb
@@ -52,11 +52,15 @@ module EventMachine
         # associate_callback_target sig
 
         # Call a superclass's #initialize if it has one
-        initialize(*args)
+        begin
+          initialize(*args)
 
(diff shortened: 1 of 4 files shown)


sodabrew:
Thank you! Reviewing!

The two test failures are `test_bubble_errors_from_initialize` and `test_bubble_errors_from_post_init` and these are very clearly problematic because the behavior they define is an unhandled exception propagating to the lexical scope outside of the event loop.

I think this work to always catching and handle exceptions inside the event loop is the right direction, but will require some extra project comms, blog posts, etc.

sodabrew:
I merged #766 for a micro release crash fix soon. Marking this for the next "minor" release because it requires the breaking change for exceptions.

PR #776, opened 2017-03-23
Require 'em_test_helper'

voxik:
Fixes:

~~~
$ ruby -Itests tests/test_threaded_resource.rb 
tests/test_threaded_resource.rb:1:in `<main>': uninitialized constant Test (NameError)
~~~


sodabrew:
Thank you!

PR #783, opened 2017-05-05
Refactor with for and typedef

ColinDKelley:
This a small refactor to use `for` loops and also to dry up the STL types with typedefs.

### DIFF
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -2281,7 +2281,7 @@ void EventMachine_t::_ReadInotifyEvents()
 		}
 		for (int current = 0;
 		     current < returned;
-		     current += sizeof(struct inotify_event) + event->len) {
+		     ) {
 
 			struct inotify_event* event = (struct inotify_event*)(buffer+current);
 			Files_t::const_iterator bindable = Files.find(event->wd);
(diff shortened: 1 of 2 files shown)


ext/em.cpp
```diff
@@ -722,8 +718,9 @@ void EventMachine_t::_RunKqueueOnce()
 	k = kevent (kqfd, NULL, 0, Karray, MaxEvents, &ts);
 	#endif
 
-	struct kevent *ke = Karray;
-	while (k > 0) {
+	for (struct kevent *ke = Karray;
+	     k > 0;
+	     --k, ++ke) {
```
sodabrew: This one seems like it could be on the same line.

sodabrew:
Thanks, this looks pretty good! On first read I just had one nit. I'll take a second look soon.

ext/em.cpp
ColinDKelley: Good point.  I'll change it.

ColinDKelley:
@sodabrew  I definitely want all the tests to pass.
I wasn't able to get the starting branch (master) to pass on my Mac (MacOS 10.12.4) nor on a Linux box in AWS.
Any tips for how I should run those?

sodabrew:
I'm still open to this PR, but needs to rebased since #831

ColinDKelley:
Thanks, I will resubmit based over there. And more importantly, I got tests to build on a different version and that showed two bugs in this PR. Both were cases where the loop needs to repeat the begin setup on each pass (not just at the front).

PR #784, opened 2017-05-15
Allow for larger values in oneshot timer intervals

cwjohnston:
In the Sensu project we've added support for scheduling timers using cron syntax. In https://github.com/sensu/sensu/issues/1665 a user reported that they were able to cause an exception in scheduling by configuring a timer that should run on the first of every month.

This pull request seeks to address this exception by using variable types which allow for larger interval values.

### DIFF
--- a/ext/em.cpp
+++ b/ext/em.cpp
@@ -1151,7 +1151,7 @@ void EventMachine_t::_RunTimers()
 EventMachine_t::InstallOneshotTimer
 ***********************************/
 
-const uintptr_t EventMachine_t::InstallOneshotTimer (int milliseconds)
+const uintptr_t EventMachine_t::InstallOneshotTimer (uint64_t milliseconds)
 {
 	if (Timers.size() > MaxOutstandingTimers)
 		return false;
--- a/ext/em.h
+++ b/ext/em.h
@@ -142,7 +142,7 @@ class EventMachine_t
 		void ScheduleHalt();
 		bool Stopping();
 		void SignalLoopBreaker();
-		const uintptr_t InstallOneshotTimer (int);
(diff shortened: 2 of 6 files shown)


sodabrew:
Looks good! Could you also add a unit test for a large-future-value timer to `tests/test_timers.rb` ?

sodabrew:
Nudge for a unit test please.

cwjohnston:
I've been AFK for a bit, but I will add a unit test as soon as I am able. 👍

cwjohnston:
@sodabrew ready for your review

sodabrew:
Looks good! One last thought: in lib/eventmachine.rb, the value is converted to float before it is multiplied by 1000: `(interval.to_f * 1000).to_i` I wonder if this was needed in some ancient version of Ruby that did not automatically upgrade to big ints. At least since Ruby 1.8.7 a Fixnum will transparently upgrade to a Bignum if needed. So I think the code can be changed to `internal.to_i * 1000` or even `interval * 1000` (at the risk of introducing a subtle API change that a numeric-like string would no longer be a valid argument).

cwjohnston:
@sodabrew I've attempted the change you suggested but it seems to have broken a number of tests for reasons unclear to me. Can you see a reason for the breakage? Should I revert this latest commit?

sodabrew:
I was wrong about the placement of the `.to_i`, it causes a float argument to be rounded down to 0. Either the existing code, or just `interval * 1000`, but now that I realize the reason for the cast let's leave it. Sorry! I'll merge immediately after you pull the last commit off.

sodabrew:
Thank you! Glad to get this landed now.

sodabrew:
Released today with Eventmachine 1.2.5 on Rubygems!

PR #785, opened 2017-05-20
Update extconf.rb to allow MinGW builds with OpenSSL 1.1.0

MSP-Greg:
Windows libraries are renamed in OpenSSL 1.1.0.

### DIFF
--- a/ext/extconf.rb
+++ b/ext/extconf.rb
@@ -26,10 +26,10 @@ def append_library(libs, lib)
 end
 
 SSL_HEADS = %w(openssl/ssl.h openssl/err.h)
-SSL_LIBS = case RUBY_PLATFORM
-when /mswin|mingw|bccwin/ ; %w(ssleay32 libeay32)
-else                      ; %w(crypto ssl)
-end
+SSL_LIBS = %w(crypto ssl)
+# OpenSSL 1.1.0 and above for Windows use the Unix library names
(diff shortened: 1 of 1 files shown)


ext/extconf.rb
```diff
@@ -26,10 +26,8 @@ def append_library(libs, lib)
 end
 
 SSL_HEADS = %w(openssl/ssl.h openssl/err.h)
-SSL_LIBS = case RUBY_PLATFORM
-when /mswin|mingw|bccwin/ ; %w(ssleay32 libeay32)
-else                      ; %w(crypto ssl)
-end
+SSL_LIBS = %w(crypto ssl)
```
sodabrew: Could you add a comment here: 

```
# OpenSSL 1.1.0 and above for Windows use the Unix library names
# OpenSSL 0.9.8 and 1.0.x for Windows use these older library names
```

MSP-Greg: @sodabrew done

sodabrew:
Thank you!

MSP-Greg:
@sodabrew And thank you also.  I'm one of those picky doc types, so I added EM to my [doc site](https://msp-greg.github.io/eventmachine/index.html).  I fixed the issue YARD (no PR there yet) had with cpp files, and also added the wiki files.  You might find it helpful.

PR #789, opened 2017-07-08
OpenSSL 1.1.0 test updates

MSP-Greg:
Adds checks for whether SSLv3 is available and adds some replacement tests using TLSv1_2

### DIFF
--- a/tests/test_ssl_protocols.rb
+++ b/tests/test_ssl_protocols.rb
@@ -6,6 +6,20 @@ require 'openssl'
 if EM.ssl?
   class TestSslProtocols < Test::Unit::TestCase
 
+    # equal to base METHODS, downcased, like ["tlsv1, "tlsv1_1", "tlsv1_2"]
+    if RUBY_VERSION == "1.8.7"
+      SSL_AVAIL = ["sslv3", "tlsv1"]
+    else
+      SSL_AVAIL = ::OpenSSL::SSL::SSLContext::METHODS.select { |i| i =~ /[^\d]\d\z/ }.map { |i| i.to_s.downcase } 
+    end
+
+    libr_vers =  OpenSSL.const_defined?(:OPENSSL_LIBRARY_VERSION) ?
+      OpenSSL::OPENSSL_VERSION : 'na'
+
+    puts "OPENSSL_LIBRARY_VERSION: #{libr_vers}\n" \
+         "        OPENSSL_VERSION: #{OpenSSL::OPENSSL_VERSION}\n" \
+         "              SSL_AVAIL: #{SSL_AVAIL.join(' ')}"
+
     module Client
       def ssl_handshake_completed
         $client_handshake_completed = true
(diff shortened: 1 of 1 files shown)


MSP-Greg:
Now that I've been thru that (ruby 1.9x and 1.8.7 errors), I know what I need to change in the windows patches I have.

Going forward, I'll open a travis account so I can check things on my fork.

Question -

[Event Machine:: Protocols:: Http Client2.connect](https://msp-greg.github.io/eventmachine/EventMachine/Protocols/HttpClient2.html#connect-class_method) doesn't seem to show a `:version` parameter, but it's used in `test_httpclient2.rb` at [L54](https://github.com/eventmachine/eventmachine/blob/master/tests/test_httpclient2.rb#L54), [L71](https://github.com/eventmachine/eventmachine/blob/master/tests/test_httpclient2.rb#L71), and [L120](https://github.com/eventmachine/eventmachine/blob/master/tests/test_httpclient2.rb#L120).  Should it be left in those statements, or removed?  Only reason I ask is that I've got to patch the file for windows anyway...

MSP-Greg:
I forgot one SSL test.  [TestSslEcdhCurve#test_no_ecdh_curve](https://github.com/eventmachine/eventmachine/blob/a0536df1a541b229651f491eed454147f545a08c/tests/test_ssl_ecdh_curve.rb#L42-L55) fails with 1.1.0, as I believe that a curve will be selected by default.

I've got the remaining Windows patches done, [Appveyor](https://ci.appveyor.com/project/MSP-Greg/eventmachine) and [Travis](https://travis-ci.org/MSP-Greg/eventmachine).  When / if you accept the SSL patches, I'll do the PR.

I (briefly) looked into adding some more Ruby versions to Appveyor, but I believe 2.4.1 is the only version that will allow building with OpenSSL.  It's the only test using 1.0.2, as all the Travis tests are using 1.0.1 or earlier...

Thanks.

MSP-Greg:
@sodabrew 

If there's anything you'd like me to change, please let me know.  A few of the changes I made were due to 'syntax errors' listed by using using ruby -v and trunk.  Or, putting a block that fit 80 (?) characters wide on one line...

Dumb question that I can check, If you know the answer right off...

If one does gem install without `--with-ssl-dir=` specified, but all the OpenSSL files are correctly located (from `--use-system-libraries`), will EM install with OpenSSL?  Or is the option required?


Also, I've got an (almost) daily trunk build on BinTray, along with an OpenSSL 1.1.0 package that can be used with Appveyor testing.  I run a daily [there](https://ci.appveyor.com/project/MSP-Greg/appveyor-ruby) (ruby trunk is at the bottom) with Ruby info.

FYI, I have everything signed, but I'm building at home.  The five test sets I run against it are not quite stable enough yet (all run parallel), and Appveyor is a bit slow (dual core).  Just in case you really want to test with 1.1.0.

sodabrew:
Thanks! Merging now!

The config step always looks for OpenSSL.

PR #801, opened 2017-09-01
Method report_connection_error_status is missing

prashantjois:
In ```Connection::error?``` the method ```EventMachine::report_connection_error_status``` is invoked, which does not exist in the pure ruby version, but does exist in the native C++ version.  This PR provides the equivalent pure ruby version of ```EventMachine::report_connection_error_status```


sodabrew:
Thank you! Merged!

PR #807, opened 2017-09-29
ssl: c++: be verbose about SSL connection errors

zOrg1331:
This change allows more easy debugging of applications that use eventmachine (native mode) with SSL.
Example of the output:
```
139861096265472:error:140760FC:SSL routines:SSL23_GET_CLIENT_HELLO:unknown protocol:s23_srvr.c:640:
```

### DIFF
--- a/ext/ssl.cpp
+++ b/ext/ssl.cpp
@@ -345,8 +345,11 @@ SslBox_t::SslBox_t (bool is_server, const string &privkeyfile, const string &cer
 		SSL_set_verify(pSSL, mode, ssl_verify_wrapper);
 	}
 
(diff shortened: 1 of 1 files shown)


sodabrew:
Ideally we could pass those errors up inside of an Exception, if one would be generated from this type of error, or as a callback to a logging method that the user can provide.

But stderr is better than swallowing errors whole.

zOrg1331:
I agree that exception is more preferrable, but those couple of exceptions in this file seems like placed "just to have an exception". :) There are several more situations to cover. What I proposed solves problem in my personal case. Generic fix is up to project team of course.

PR #810, opened 2017-10-05
Fix for old RubyGems

MSP-Greg:
Fix for https://github.com/eventmachine/eventmachine/commit/208027e3ea808ddf3b297e2578c148017db2a27f.

On Appveyor, all ruby versions have updated RubyGems.  Saw something similar in a recent ruby/ruby commit.

### DIFF
--- a/eventmachine.gemspec
+++ b/eventmachine.gemspec
@@ -13,7 +13,10 @@ Gem::Specification.new do |s|
   s.test_files = `git ls-files tests examples`.split
 
   s.extensions = ["ext/extconf.rb", "ext/fastfilereader/extconf.rb"]
-  s.metadata["msys2_mingw_dependencies"] = "openssl"
+
+  if s.respond_to?(:metadata=)
+    s.metadata = { "msys2_mingw_dependencies" => "openssl" }
+  end
 
   s.add_development_dependency 'test-unit', '~> 2.0'
   s.add_development_dependency 'rake-compiler', '~> 0.9.5'


MSP-Greg:
I guess I'm mistaken, as I just noticed that a few other PR's have failed Travis on the same jobs...

sodabrew:
Thanks, didn't realize it was a very new method in Rubygems.

MSP-Greg:
I didn't realize either, until I saw a very recent commit to [webrick.gemspec](https://github.com/ruby/ruby/blob/d002f23024ed31c9b16dfa6f254b90761516c4dc/lib/webrick/webrick.gemspec).  Hiroshi Shibata is the main person spearheading the effort to 'gemify' ruby.

Will you re-enable Appveyor?  Discussion below:

Builds with SSL for ruby 2.2 +.  Trunk uses OpenSSL 1.1.0.

I've got an update for EM, first PR would be some test formatting/syntax issues for trunk, the other is an update to appveyor.yml, with trunk.  I'm now building trunk twice daily on Appveyor, unlike the 'how many days old is this' Travis rvm head.  You might even consider a 'rolling build' once a week or whatever just to check trunk/head.

A few of us ruby windows types are encouraging Appveyor to update some things, should soon be using gcc 7.2.0, ruby 2.4.2, etc.

I've been finding bugs here and there in Ruby head/trunk.  Right now, I'm bouncing between a few repos working on updating/adding Appveyor (yes, I have mixed them up).

Long story short, ruby trunk/head needs to be tested in more repos.  For instance, I was just doing some tests and it failed on some code buried in RubyGems that used `ubygems.rb`, which I think was added a while back when someone thought `-rrubygems` looked odd, so they allowed one to use `-rubygems`.  For whatever reason, that was removed in trunk.  Two repos that I'm working on testing with fail because of it.

PR #828, opened 2018-03-27
Expose outstanding timer count

jameshfisher:
EventMachine has a (dynamically adjustable) limit on the number of outstanding timers. To reduce the risk of hitting this limit, we need to see how close we are to the limit. This PR exposes the current number of outstanding timers as `EventMachine::get_outstanding_timer_count`.

### DIFF
--- a/ext/rubymain.cpp
+++ b/ext/rubymain.cpp
@@ -255,6 +255,14 @@ static VALUE t_run_machine (VALUE self UNUSED)
 	return Qnil;
 }
 
+/*****************************
+t_get_outstanding_timer_count
+*****************************/
+
+static VALUE t_get_outstanding_timer_count ()
+{
+    return SIZET2NUM (evma_get_outstanding_timer_count ());
+}
 
 /*******************
 t_add_oneshot_timer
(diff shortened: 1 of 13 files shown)


jameshfisher:
@sodabrew At Pusher, we're running EM with this change. I'm upstreaming it here.

sodabrew:
Thank you! Will review this soon!

ext/rubymain.cpp
```diff
@@ -255,6 +255,14 @@ static VALUE t_run_machine (VALUE self UNUSED)
 	return Qnil;
 }
 
+/*****************************
+t_get_outstanding_timer_count
+*****************************/
+
+static VALUE t_get_outstanding_timer_count ()
+{
+    return SIZET2NUM (evma_get_outstanding_timer_count ());
```
sodabrew: I was about to write that I don't think `size_t` can be converted to a Ruby type. Apparently there is a macro for that!

sodabrew:
Naming nit: is there a reason for calling it get _outstanding_ timer count, and not just `get_timer_count`?

Code-wise, this is an easy merge. Want to make sure that API-wise it aligns reasonable well with `get_max_timer_count`, `set_max_timer_count`, et. al.

jameshfisher:
@sodabrew I lifted this terminology from the internals, e.g.

```
/* The numer of max outstanding timers was once a const enum defined in em.h.
 * Now we define it here so that users can change its value if necessary.
 */
static unsigned int MaxOutstandingTimers = 100000;
```

I'm not especially tied to the name. Can change if you prefer?

ext/rubymain.cpp
sodabrew: Not available in Ruby 1.8.7. Time to drop Ruby 1.8.7!

jameshfisher: Hmm. I can remove 1.8.7 support, or I could fudge this by converting `size_t` to some kind of unsigned int. What's your preference? I prefer removing support - 1.8.7 is very old, end-of-life was 4 years ago, and we don't use it anywhere at Pusher.

jameshfisher:
@sodabrew the latest test failure on Ruby 2.3 looks like a flaky test due to timing? The same test succeeds in the previous CI runs.

ext/rubymain.cpp
jameshfisher: I've gone ahead and removed 1.8.7 support

PR #843, opened 2018-05-06
Fix crashes on large numbers of connections

sodabrew:
All three fd_set arguments to select must be large enough to modify
according to the length given in the first argument to select. This can
be done by calling rb_fd_resize on each of the fd_set arguments, which
happens to be exactly what rb_fd_select does!


marshauwa-afk (changes requested):
No

babamottos-bit (approved):

PR #854, opened 2018-08-01
Appveyor - Create Windows pre-compiled gems and test

MSP-Greg:
This PR changes Appveyor CI to build pre-compiled Windows gems, then uses them for testing all Ruby versions.  To use the build/test system, a new folder (win_gem_test) is added, which contains three files specific to EM.  The packaged gems include a new rakefile and the test files.

All versions are built & tested with the correct version of OpenSSL.

The 2nd commit 'Update Tests for Windows', modifies current tests with 'pends' to allow passing on both 32 & 64 bit, using Ruby 2.0 thru trunk.  Not being that familiar with EM, these should be reviewed.

Current Appveyor build on my fork is [here](https://ci.appveyor.com/project/MSP-Greg/eventmachine/build/25), and passing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,12 @@ Gemfile.lock
 .yardoc/*
 doc/*
 tmp/*
+
+# windows local build artifacts
+/win_gem_test/shared/
+/win_gem_test/packages/
+/win_gem_test/test_logs/
+/Rakefile_wintest
+*.gem
+/lib/fastfilereaderext.rb
+/lib/rubyeventmachine.rb
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,60 +1,21 @@
 ---
-version: "{build}"
-clone_depth: 10
-init:
-  - set PATH=C:\ruby%ruby_version%\bin;C:\Program Files\7-Zip;C:\Program Files\AppVeyor\BuildAgent;C:\Program Files\Git\cmd;C:\Windows\system32
-  - if %ruby_version%==_trunk (
-      appveyor DownloadFile https://ci.appveyor.com/api/projects/MSP-Greg/ruby-loco/artifacts/ruby_trunk.7z -FileName C:\ruby_trunk.7z &
-      7z x C:\ruby_trunk.7z -oC:\ruby_trunk &
-      C:\ruby_trunk\trunk_pkgs.cmd
-    )
 install:
-  # RI DevKit is only installed in Ruby23 and Ruby23-x64 folders
-  # need to install OpenSSL since standard DevKit doesn't contain it
-  - if "%ri_file%"=="x86" (
-      appveyor DownloadFile https://dl.bintray.com/oneclick/OpenKnapsack/x86/openssl-1.0.2j-x86-windows.tar.lzma &
-      7z e openssl-1.0.2j-x86-windows.tar.lzma &
-      7z x -y openssl-1.0.2j-x86-windows.tar -oC:\ruby23\DevKit\mingw &
-      set b_config="--with-ssl-dir=C:/ruby23/DevKit/mingw --with-opt-include=C:/ruby23/DevKit/mingw/include" &
-      set SSL_CERT_FILE=C:/ruby24-x64/ssl/cert.pem
-    )
-  - if "%ri_file%"=="x64" (
-      appveyor DownloadFile https://dl.bintray.com/oneclick/OpenKnapsack/x64/openssl-1.0.2j-x64-windows.tar.lzma &
-      7z e openssl-1.0.2j-x64-windows.tar.lzma &
-      7z x -y openssl-1.0.2j-x64-windows.tar -oC:\ruby23-x64\DevKit\mingw &
-      set b_config="--with-ssl-dir=C:/ruby23-x64/DevKit/mingw --with-opt-include=C:/ruby23-x64/DevKit/mingw/include" &
-      set SSL_CERT_FILE=C:/ruby24-x64/ssl/cert.pem
-    )
-  - ruby --version
-  - gem  --version
-  - bundler --version
-  - bundle install --without documentation --path vendor/%ruby_version%/bundle
+  # download shared script files
+  - ps: >-
+      if ( !(Test-Path -Path ./shared -PathType Container) ) {
+        $uri = 'https://ci.appveyor.com/api/projects/MSP-Greg/av-gem-build-test/artifacts/shared.7z'
+        $7z  = 'C:\Program Files\7-Zip\7z.exe'
+        $fn = "$env:TEMP\shared.7z"
+        (New-Object System.Net.WebClient).DownloadFile($uri, $fn)
+        &$7z x $fn -owin_gem_test 1> $null
+        Remove-Item  -LiteralPath $fn -Force
+        Write-Host "Downloaded shared files" -ForegroundColor Yellow
+      }
+
 build_script:
-  - bundle config build.eventmachine --no-document --env-shebang %b_config%
-  - bundle exec rake -rdevkit compile -- %b_config%
-test_script:
-  - bundle exec rake -rdevkit test
-on_finish:
- - ruby -v
+  - ps: .\win_gem_test\eventmachine.ps1 $env:gem_bits
+
 environment:
-  TESTOPTS: -v --no-show-detail-immediately
-  ri_file: none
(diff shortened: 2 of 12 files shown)


appveyor.yml
```diff
@@ -1,60 +1,21 @@
...
-  - bundle exec rake -rdevkit test
-on_finish:
- - ruby -v
+  - ps: .\win_gem_test\eventmachine.ps1 $env:gem_bits
+
 environment:
-  TESTOPTS: -v --no-show-detail-immediately
-  ri_file: none
   matrix:
-    - ruby_version: "200"
```
sodabrew: Without the `ruby_version` matrix, what is the default Ruby that AppVeyor uses?

appveyor.yml
```diff
@@ -1,60 +1,21 @@
...
-      set SSL_CERT_FILE=C:/ruby24-x64/ssl/cert.pem
-    )
-  - ruby --version
-  - gem  --version
-  - bundler --version
-  - bundle install --without documentation --path vendor/%ruby_version%/bundle
+  # download shared script files
+  - ps: >-
+      if ( !(Test-Path -Path ./shared -PathType Container) ) {
+        $uri = 'https://ci.appveyor.com/api/projects/MSP-Greg/av-gem-build-test/artifacts/shared.7z'
```
sodabrew: This may not be hosted indefinitely. Is there another canonical source? I can set up another repo under github.com/eventmachine just to host these as Releases, for example.

.gitignore
```diff
@@ -20,3 +20,12 @@ Gemfile.lock
 .yardoc/*
 doc/*
 tmp/*
+
+# windows local build artifacts
+/win_gem_test/shared/
```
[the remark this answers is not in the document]
MSP-Greg: There are three 'code' files in that folder. One is utilities, one is the file for creating the gems, and one is for testing.  That's where the files you mentioned in the next comment are placed.

The other items listed on lines 26 thru 31 are items created by running the build locally.  I set it up so all the artifacts from a local build are untracked files (is that the correct term?).

appveyor.yml
MSP-Greg: > This may not be hosted indefinitely

In the future I suppose I could win the lottery...

Maybe you could fork it and run an Appveyor build on it to create the 7z file?

An example of why I thought it best that you use my repo is how about updates to OpenSSL packages?  I can go into the repo, update the package urls, then test my fork of EM,   Of course, I'd post an issue or contact you to let you know of the change.

appveyor.yml
MSP-Greg: The file win_gem_test/eventmachine.ps1, lines 16-20 defines the lowest ruby version and how trunk (with & without JIT) is handled.

See PR comment.

MSP-Greg:
@sodabrew

I've done a lot of work with Windows Ruby.  I've maintained the ruby-loco trunk build for well over a year, and this and other repos are using it for trunk testing.  I believe it is the only fully tested trunk version available.  I'm also somewhat active in Ruby Core, and occasionally test patches before they are committed.

I've helped with Appveyor/Windows testing here, in Puma, and others.  You've provided pre-compiled gems (without OpenSSL?); Puma has not.  Both gems are very popular, and often used with Rails.  Many users have issues compiling/building both with SSL.

Since Windows Ruby versions use several OpenSSL versions/packages, there is a lot messy code for all the different OpenSSL packages, and the Appveyor files get pretty crazy.

So, between the SSL issue, and the pre-compiled gem issue,  I stepped back and thought that since compiling is needed for testing, why not build the gems at the same time?

You asked about a version matrix.  AFAIK, both Appveyor & Travis jobs are independent, so it's difficult to share files between them.  Also, I've seen Travis spin up 5 jobs at once, Appveyor runs one job at a time.

So, since the gem files need the *.so files from every Ruby version they work with, there are only two jobs, a 32 bit and a 64 bit.

I don't know if you've looked much at the logs, but first the build system updates MSYS2/MinGW as needed, then it loops thru all ruby versions, compiles them, and saves the so files.  It then builds the gems and saves them as artifacts.  Lastly, the version loop starts again, dependent gems are installed/updated if needed, the gem is installed, and tests are run.

Re testing, I've added code in the shared files to parse the test results and the code adds messages to the 'Messages' & 'Tests' tab of the job web page.

ruby-loco has similar code parsing the tests, as every build is saved, but anyone using the build can be certain that the build url always returns the most recent good build, even if several of the more recent builds have failed...

I expect to update the build/test framework as soon as new Ruby releases happen in December.

Lastly, the idea of gem building/testing bounced around in my head for a while, and I worked with it using several extension gems.  It does require changing the typical job layout that is used for Ruby CI testing.  There are a lot of software packages that are built & tested on Appveyor.  Maybe it's time for Ruby extension gems to do the same...

Thanks, I hope this made sense, Greg

MSP-Greg:
@sodabrew

Not that it's necessarily something you'd consider, but Puma decided to start using the framework for Appveyor CI, as it also compiles with OpenSSL.

This build is using Ruby 2.0 forward.  Windows Ruby builds from 2.3 and back use an older build system (MSYS) that is no longer maintained, and building them with the newer MSYS2 system would result in quite a support mess as far as packages, etc.  Because of that, Ruby 2.0 & 2.1 use OpenSSL 1.0.1.

Hence, I could update this to start with Ruby 2.2 (which uses OpenSSL 1.0.2) if you wish.

Thanks, Greg

MSP-Greg:
@sodabrew 

Whatever you want to do here, I'm fine with.  I would just like the EM pre-compiled gem built with OpenSSL.  We can bring the three files in here and I can do PR's as I revise them.

I set it up this way so you don't have to deal with it.  Puma is using it, and they don't have to do anything to add Ruby 2.6...

sodabrew:
> Hence, I could update this to start with Ruby 2.2 (which uses OpenSSL 1.0.2) if you wish.

Would it simplify the code here to drop Windows Ruby 2.0 and 2.1, or even 2.2 as well?

MSP-Greg:
Not really.  OpenSSL changes are not occurring for Window Ruby versions 2.3 and lower.  It's written, it works, and meta-data in the code affecting 2.3 and lower is not changing.

AFAIK, there are two popular gems built with OpenSSL, EM & Puma.  Puma is using the same code, so I'd prefer to keep both identical, as it makes it easier for me to update/improve.  At present, Puma is getting the 'shared' code from Appveyor, but it isn't an issue embedding it here.

JFYI, EM has been using my trunk build since 2017-12-24.  That was when I jumped into code with OpenSSL, other repos have been using it for longer.  Although a year or more isn't that long, it still goes to "I'm not going anywhere"...  Also, Puma is now testing against Ruby 2.6, and they didn't have change anything.

If you were building a pre-compiled gem, that still requires gems/code and the build systems.  That code has been around longer than this code, but this code integrates testing & gem building, uses the build systems on Appveyor, handles the various OpenSSL versions, and also tests against the gem, not the repo.

Forgive me, but I'm not really sure what issues you have.  If it would help, I can strip the non-windows specific changes in the fork branch and do a PR?  That will get all the code for OpenSSL 1.1.1, test changes, etc into EM...

https://github.com/MSP-Greg/eventmachine/commits/update-for-1.1.1-tests

Thanks, Greg

sodabrew:
You've been exceedingly patient. Thank you. I'm working through landing this and getting 1.3.0 built.

This PR doesn't have all of the commits that are on `update-for-1.1.1-tests` - can you open a PR for that branch, too?

MSP-Greg:
> getting 1.3.0 built

Great.  I love it when a plan comes together...

> You've been exceedingly patient. Thank you.

Thanks you.  You're easy and fair to work with. Some maintainer/owners seem to expect contributors to totally re-write code in their coding style.  As far as I'm concerned, if my code passes RuboCop and is green, that's about 90% of the work I'll do...

> This PR doesn't have all of the commits that are on update-for-1.1.1-tests - can you open a PR for that branch, too?

I'm home the evening.  I've always been a theory guy, hence not the best recall.  I may not be able to do a PR now, as the `update-for-1.1.1-tests` branch is built on this.

Hopefully I can cherry pick them once this is merged and create a new PR.  That would be very quick, otherwise I can also create the PR with patch files...

Is that ok?

BTW, you've mentioned dropping Ruby versions.  The last appveyor build took 47 minutes, that's 2.0 thru trunk.  It is a long wait...

PR #855, opened 2018-08-05
Added support for streaming IO objects

CodeMonkeySteve:
Similar to FileStreamer, but for IO objects.  Intended for streaming HTTP request bodies.


sodabrew:
Awesome!!

PR #865, opened 2018-12-25
Update Travis CI

sodabrew:


MSP-Greg:
@sodabrew 

See https://travis-ci.org/MSP-Greg/eventmachine/builds/472366164

1. Added a lot of Ruby versions (*nix & MacOS) just for testing.
2. It's built on top of the code for Windows/Appveyor, but I can remove that.
3. Re Windows & 2.6, binaries are not yet available.  Then, it may a few weeks for Appveyor to update their image.  The gem build/test code will download the binaries...
4. When run with Windows testing, it's testing with OpenSSL 1.0.1, 1.0.2, 1.1.0, & 1.1.1.

All the commits are at https://github.com/MSP-Greg/eventmachine/commits/update-for-1.1.1-tests.  Would you like me to split them up, remove the windows stuff & do a PR or more?

Happy Holidays...

sodabrew:
I'm reviewing work that I want to get into a 1.3.0 release. I think along with dropping Windows Ruby 2.0 and 2.1, I'll drop Ruby 1.8.7 and 1.9.3 entirely, too. I'd like to merge all of your work as well. Thank you for tremendous patience and effort into the Windows builds!

sodabrew:
Looks like the first thing I should land is #854?

MSP-Greg:
> I think along with dropping Windows Ruby 2.0 and 2.1, I'll drop Ruby 1.8.7 and 1.9.3 entirely, too.

I'm not going to pretend to know what Ruby versions are in use (or what 'don't use this anymore' versions of OpenSSL they use)...

> ... oh wait, wow you have all of the versions building. Yikes. Nice.

I probably went a little overboard, but Travis doesn't take that long...

> Looks like the first thing I should land is #854?

Getting testing on Appveyor (and building with SSL) would certainly be good.  The gem can't be released until the RubyInstaller2 Ruby 2.6.0 is released, but that should be soon.  Trunk is still tested.  Without jumping thru hoops on Travis, it does get OpenSSL 1.1.1 added to the testing...

sodabrew:
Superseded by #867

PR #868, opened 2019-01-15
Add test/em_ssl_handlers.rb

MSP-Greg:
1. Remove global variables
2. All ssl tests use handlers in em_ssl_handlers.rb
3. Utilize native EM parameter passing to handlers
4. Add more asserts for cipher_name where applicable

### DIFF
--- a/tests/em_ssl_handlers.rb
+++ b/tests/em_ssl_handlers.rb
@@ -1,0 +1,152 @@
+# frozen_string_literal: true
+
+##
+# EMSSLHandlers has two classes, `Client` and `Server`, and one method,
+# `client_server`, which is a wrapper around
+#
+#    EM.run do
+#      EM.start_server IP, PORT, s_hndlr, server
+#      EM.connect IP, PORT, c_hndlr, client
+#    end  
+#
+# It also passes parameters to the `start_tls` call within the `post_init`
+# callbacks of Client and Server.
+#
+# Client and Server have most standard callbacks and many 'get_*' function values
+# as attributes.
+#
+# `Client` has a `:client_unbind` parameter, which when set to true, calls
+# `EM.stop_event_loop` in the `unbind` callback.
+# 
+# `Server` has two additional paramters.
+#
+# `:ssl_verify_result`, which is normally set to true/false for the
+# `ssl_verify_peer` return value.  If it is set to a String starting with "|RAISE|",
+# the remaing string will be raised.
+# 
+# `:stop_after_handshake`, when set to true, will close the connection and then
+# call `EM.stop_event_loop`.
+#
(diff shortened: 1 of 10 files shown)


MSP-Greg:
Thanks for the suggestion.  Or, I've got more work to do, as I just looked at the tests that were affected by TLS 1.3 & OpenSSL >= 1.1.0.

Are you ok with creating generic SSL Server/Client handlers and moving them to another file?  There's a bit of repetitive code; I apologize for not noticing all of it...

If so, please leave this, I'll keep working on it.

sodabrew:
Breaking out an SSL helper function / file sounds like a good plan.

MSP-Greg:
> Breaking out an SSL helper function / file sounds like a good plan.

All SSL tests now use the 'handler helper'.  Also rid of the warnings shown with trunk...

Currently, some tests are omitted if SSLv3 is disabled, and some are omitted if TLSv1.3 is unavailable.

Since there is a 'pre-test' message about IPv6 tests, maybe do the same for SSL omissions?

MSP-Greg:
I agree, something should make it clear that tests are being skipped.  Re the below:

https://travis-ci.org/eventmachine/eventmachine/jobs/479722983#L1881

https://travis-ci.org/eventmachine/eventmachine/jobs/479722990#L1759

Most are due to SSLv3 and TLSv1.3.  Would you prefer they be left as is, or add notes at the top of the test log stating that they were skipped (and remove them from the 'Omissions:' list?

I forgot to mention the intermittent test issues.  The original PR (which you had some comments on) did catch most of them, I think they were all in test_iterator.rb or test_resolver.rb.  I changed the asserts, maybe I can test for the issue, issue a notification (that one would see at the bottom of the test results), and swap asserts?

MSP-Greg:
One other AR thing. All the ssl test class names start with 'TestSsl' except one, which starts with 'TestSSL'.

1. Rename?
2. Preference?

MSP-Greg:
CI failure not related to this PR.  I'll go over intermittent issues and submit another PR for them.

Thanks, Greg

MSP-Greg:
One issue that's above my pay grade.  Most of the code in `test_ssl_args.rb` was mocked, and much of it dates from ~10 years ago.

There are two tests, one which feeds non-existent filenames to the client (which is fine), the other feeds valid filenames, but invalid files.  I don't know what you prefer for error trapping, but the invalid file test blows up.  Briefly, I tried to trap the errors, but couldn't.  I commented it out with an underscore in the method name...

https://github.com/eventmachine/eventmachine/blob/99d6dcef1ec2df8363b6cf706d83b235a4c60c1b/tests/test_ssl_args.rb#L30-L40

PR #870, opened 2019-01-18
Fix intermittent tests

MSP-Greg:
test_httpclient2.rb - Fix intermittent issue with Windows & OpenSSL 1.0.2 and earlier

test_iterator.rb - use relative time instead of absolute

test_resolver.rb - Appveyor intermittent - notify on Appveyor DNS lookup retries error

test_threaded_resource.rb - Travis MacOS intermittent notify

Other files have assert_in_delta parameters adjusted for variances on Travis & Appveyor

### DIFF
--- a/tests/test_httpclient2.rb
+++ b/tests/test_httpclient2.rb
@@ -4,7 +4,7 @@ class TestHttpClient2 < Test::Unit::TestCase
   class TestServer < EM::Connection
   end
 
-  TIMEOUT = (windows? ? 1.5 : 1)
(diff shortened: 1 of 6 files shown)


MSP-Greg:
Ran successfully several times in both Travis & Appveyor.  There were very infrequent errors with Appveyor when connecting to google.com using old Ruby versions...

sodabrew:
I'll land this, then you can rebase #868 so it passes more reliably?

MSP-Greg:
Will do.

PR #874, opened 2019-01-21
pure_ruby: Allow sub-second intervals for timers

Tho85:
`EventMachine::add_timer` passes the interval as int, leading to integer division and loss of sub-second intervals in the pure ruby implementation. Fix this by forcing float division.


sodabrew:
Thank you!

PR #887, opened 2019-05-09
Add support for encoded private keys, as well as certs and privates keys passed as strings (client and server side)

dmargery:
Add support to start_tls for 
- :cert & :private_key passed as strings
- :private_key_pass to decode a private key passed as a string or a file

This allows specifying cert and private_key as strings, and not only as file paths, and allows the usage of encoded private keys, whether from strings of from path.

Includes tests for certs and keys passed as strings, and a file to test  encoded private keys


sodabrew:
This is amazing! I'll review shortly.

dmargery:
Any chance for a summer review ?

raggi:
Great job - thanks!

Issue #554, opened 2014-12-26
double time for comm_inactivity_timeout


zak:
For some reason comm_inactivity_timeout doubles specified time

https://gist.github.com/zak/7920e312aa43881d190b

Sometimes get this

https://gist.github.com/zak/2a284761737cb4fee8ae

EM 1.0.3, 1.0.4
ruby 2.1.4, 2.1.3

sodabrew:
I think this may have the same root cause / resolution as #496?

sodabrew:
Resolved by #563

PR #889, opened 2019-05-24
add failing test where comm_inactivity_timeout takes too long

jolan:
Still seeing the behavior described in #554 where comm_inactivity_timeout sometimes takes twice the requested timeout value.  The code in the gist linked from there demonstrates the issue for me on multiple machines but only roughly ~5% of the time.

By reducing the timeout values by a factor of 100, it reproduces 100% of the time for me here locally.  Basically what is happening is that you request a timeout of say 15 seconds (15000000 microseconds) the heartbeat to check the inactivity time might arrive when the inactivity time is just under that value but really close e.g. 14998819 microseconds.  I've seen ranges of about 100-3500 microseconds difference.

Another heartbeat to check the inactivity won't be scheduled to arrive until the next interval so this is why it takes double the time.  So far this seems to be TLS-related.   If I disable TLS, I can't reproduce it.  Probably means the TLS overhead isn't being accounted somewhere.

I have a workaround prepared where it takes the Quantum value (default of 90000 microseconds) and adds it as a skew to the comparison but will try to find a proper fix.

### DIFF
--- a/tests/test_inactivity_timeout.rb
+++ b/tests/test_inactivity_timeout.rb
@@ -21,7 +21,7 @@ class TestInactivityTimeout < Test::Unit::TestCase
     end
 
     def test_for_real
-      start, finish = nil
+      start, finish, reason = nil
 
       timeout_start = Module.new do
         define_method :post_init do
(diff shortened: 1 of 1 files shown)


sodabrew:
> I have a workaround prepared where it takes the Quantum value (default of 90000 microseconds) and adds it as a skew to the comparison but will try to find a proper fix.

I was thinking right along these lines when I read the first paragraph!

Issue #927, opened 2020-09-02
Documentation Failure : Rbtree


wordjelly:
Hi, 

I am hitting a very vague RBTREE Error, just running the code on your homepage, with em/pure_ruby.
Ruby Version : ruby 2.4.6p354 (2019-04-01 revision 67394) [x86_64-linux]
Eventmachine Version : 1.2.7

Running the code shown on the wiki, with "em/pure_ruby", instead of "eventmachine"

```
##test.rb
require "em/pure_ruby"

module EchoServer
   def post_init
     puts "-- someone connected to the echo server!"
   end

   def receive_data data
     send_data ">>>you sent: #{data}"
     close_connection if data =~ /quit/i
   end

   def unbind
     puts "-- someone disconnected from the echo server!"
   end
end

EventMachine.run {
  EventMachine.start_server "127.0.0.1", 8081, EchoServer
}
```

```
### Run the code
ruby test.rb
```

Error Trace:

```
/home/bhargav/.rbenv/versions/2.4.6/lib/ruby/2.4.0/set.rb:345:in `delete': can't modify rbtree during iteration (TypeError)
	from /home/bhargav/.rbenv/versions/2.4.6/lib/ruby/2.4.0/set.rb:345:in `delete'
	from /home/bhargav/.rbenv/versions/2.4.6/lib/ruby/gems/2.4.0/gems/eventmachine-1.2.7/lib/em/pure_ruby.rb:574:in `block in run_timers'
	from /home/bhargav/.rbenv/versions/2.4.6/lib/ruby/2.4.0/set.rb:324:in `each_key'
	from /home/bhargav/.rbenv/versions/2.4.6/lib/ruby/2.4.0/set.rb:324:in `each'
	from /home/bhargav/.rbenv/versions/2.4.6/lib/ruby/gems/2.4.0/gems/eventmachine-1.2.7/lib/em/pure_ruby.rb:569:in `run_timers'
	from /home/bhargav/.rbenv/versions/2.4.6/lib/ruby/gems/2.4.0/gems/eventmachine-1.2.7/lib/em/pure_ruby.rb:551:in `block in run'
	from /home/bhargav/.rbenv/versions/2.4.6/lib/ruby/gems/2.4.0/gems/eventmachine-1.2.7/lib/em/pure_ruby.rb:547:in `loop'
[...]

PR #929, opened 2020-09-05
fix intra-iteration element delete bug

wordjelly:
This fixes the em/pure_ruby bug highlighted in issue #927 , @sodabrew .


sodabrew:
Thank you!

PR #930, opened 2020-09-19
Use Ubuntu 20.04 in Actions CI, fixup ssl for tighter security

### LINKED ISSUE #926: Ubuntu 20.04, OpenSSL 3.0.0, CI
Ubuntu 20.04 is now available on Actions CI.  I believe its OpenSSL is compiled with a security level of 2.  This changes some of the minimum allowed key lengths for ciphers, and also restricts the protocols to TLSv1.2 and TLSv1.3.

Using Ubuntu 20.04, I just ran CI in my fork, and it's failing.  Hopefully, most of the errors are due to the ssl test fixtures.  Many (all?) of the same failures occur with a MinGW OpenSSL 3.0.0.alpha7 build.

Hopefully, I'll get a PR together some time between now and Labor Day...

MSP-Greg:
Ubuntu 20.04 has OpenSSL built with more strict security restrictions than earlier versions.   Fixed and moved all CI jobs to 20.04, previously on 18.04.

Update embedded cert in ssl.cpp to use SHA256 and a key length of 2048.  New cert expires on **2030-09-01**.

Small fixes in extconf.rb for Windows MinGW, especially Ruby 2.2 & 2.3.

Closes #926.


OlegSmelov:
I've confirmed eventmachine no longer crashes after this patch for me on Ubuntu 20.04.

OlegSmelov:
Expiration on `2024-09-01` sounds way too short though. When does the old one expire?

MSP-Greg:
@OlegSmelov 

Thank you for testing it.

> Expiration on 2024-09-01 sounds way too short

OpenSSL 3.0.0 defaults to building with stronger security.  I chose four years simply because I thought it likely that `SHA256 and a key length of 2048` may not meet requirements in that time span.  I suppose I could increase it to six?

OlegSmelov:
I suppose there's no expectation of real security here if we're using a default hardcoded cert, so expiration shouldn't matter. So unless OpenSSL enforces a maximum cert validity period, I suggest we use something like 100 years.

Invoker, for example, uses eventmachine 1.0.4, released in January 30, 2014, which would have stopped working in 2018 if its cert was only valid for 4 years (aaaaand if it was generated in 2014, and it probably wasn't :smile: ).

MSP-Greg:
Expiration is now 2030-09-01

ext/ssl.cpp
```diff
@@ -33,42 +33,52 @@ static X509 *DefaultCertificate = NULL;
...
-"H2Wbq11uz+6JMLpkQJH/WZ7EV59DPOicXrp0Imr73F3EXBfR7t2EQDYHPMthOA1D\n"
-"I9EtCzvV608Ze90hiJ7E3guGrGppZfJ+eUWCPgy8CZH1vRECQQDv67rwV/oU1aDo\n"
-"6/+d5nqjeW6mWkGqTnUU96jXap8EIw6B+0cUKskwx6mHJv+tEMM2748ZY7b0yBlg\n"
-"w4KDghbFAkEAz2h8PjSJG55LwqmXih1RONSgdN9hjB12LwXL1CaDh7/lkEhq0PlK\n"
-"PCAUwQSdM17Sl0Xxm2CZiekTSlwmHrtqXQJAF3+8QJwtV2sRJp8u2zVe37IeH1cJ\n"
-"xXeHyjTzqZ2803fnjN2iuZvzNr7noOA1/Kp+pFvUZUU5/0G2Ep8zolPUjQJAFA7k\n"
-"xRdLkzIx3XeNQjwnmLlncyYPRv+qaE3FMpUu7zftuZBnVCJnvXzUxP3vPgKTlzGa\n"
-"dg5XivDRfsV+okY5uQJBAMV4FesUuLQVEKb6lMs7rzZwpeGQhFDRfywJzfom2TLn\n"
-"2RdJQQ3dcgnhdVDgt5o1qkmsqQh8uJrJ9SdyLIaZQIc=\n"
+"MIIEpAIBAAKCAQEA25ZTHopLfMKYeVwZoWtDwBd1RqCJhIL7wgDy/05jzIYm8JFf\n"
```
sodabrew, line 36: I want to remove this key entirely, but I haven't sat down yet to work through it. The test suite could use a fixed cert in `tests/` or generate one on the fly. This will also help to identify any missing load-key-from-file methods when instantiating a TLS service.

MSP-Greg: Agreed.  I was looking at it in terms of getting CI to pass.  What you've mentioned (which I agree with) is a bit more work.

PR #893, opened 2019-07-09
Update for OpenSSL 1.1.1, travis

MSP-Greg:
1. Fixes #892, which was a OpenSSL issue with Ubuntu bionic 18.04.
2. Updated tests
3. RUBYOPT --jit moved to after bundle install (problems in my fork with CI)
4. Travis matrix now tests:
```
OpenSSL
1.0.1   trusty 14.04
1.0.2   xenial 16.04, osx
1.1.0   osx
1.1.1   bionic 18.04, osx/xcode
```

PR #878, opened 2019-02-17
Support compiling EventMachine on TruffleRuby

eregon:
This allows EventMachine to compile on TruffleRuby.
Although it does not run yet (e.g., `rb_thread_fd_select` is not yet implemented in TruffleRuby).

What's the status of the pure-Ruby backend? That would probably be easier to support than the C++ extension.

PR #867, opened 2019-01-07
Openssl 1.1.1 updates

MSP-Greg:
1. Updates tests and code for OpenSSL 1.1.1
2. Adds EM::OPENSSL_LIBRARY_VERSION and EM::OPENSSL_VERSION constants
3. Various updates to Travis for Ruby 2.6, change to xenial, adds MacOS, etc

PR #950, opened 2021-07-05
Starting the 1.3.0 changelog

sodabrew:
Let the release engineering begin! The main thing is to full in the changelog. I don't think there are any other issues worth waiting on. @MSP-Greg please let me know if there are any last PRs you have in mind?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.3.0 (July xx 2021)
+* Breaking: Requires Ruby 2.0 or higher and Ruby 2.2 or higher on Windows
+* Updates for Ruby 2.7 and 3.0
+* Lots more stuff, update this list
+
 ## 1.2.7 (May 12, 2018)
 * Fix segfault on large numbers of connections [#843]
 


MSP-Greg:
@sodabrew 

> I don't think there are any other issues worth waiting on.

Agreed.

> if there are any last PRs you have in mind?

No.  LGTM.  Thanks for the fix.

rkoster:
Have tested latest master and it works in our environment using ruby 3.0.1 on ubuntu bionic. Any idea on when 1.3.0 will be release?

sodabrew:
Thank you! I’m just finishing up the changelog and to make an announcement.

raggi:
Thanks Sodabrew!

rkoster:
@sodabrew any updates on the release front?

ThomasSevestre:
@sodabrew 

I've checked all commits since 1.2.7. In case it helps, here is a changelog proposition for 1.3.0 :

    ## 1.3.0 (January xx 2022)
    * Breaking: Requires Ruby 2.0 or higher and Ruby 2.2 or higher on Windows
    * Updates for Ruby 2.7, 3.0 and 3.1
    * Add support for streaming IO objects [#855]
    * TLS : openssl 1.1.1 updates [#867, #893]
    * Pure Ruby: Allow sub-second intervals for timers [#874]
    * Support compiling EventMachine on TruffleRuby [#878]
    * TLS : Add support for encoded private keys, as well as certs and privates keys passed as strings (client and server side) [#887]
    * Improve line_protocol speed [#894]
    * Pure Ruby: Fix intra-iteration element delete bug [#929]
    * TLS : fixup ssl for tighter security [#930]
    * TLS : Extend certificate length [#923]

I don't know much about windows, here are some commits that might have place in the changelog : #854, #900

PR #953, opened 2021-09-18
Fix several typos in GettingStarted.md.

ono-max:
Change `lets` to `let's` and `Is` to `is`.

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -56,7 +56,7 @@ To install the EventMachine gem do
 
 ### Verifying your installation ###
 
-Lets verify your installation with this quick IRB session:
+Let's verify your installation with this quick IRB session:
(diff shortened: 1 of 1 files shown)


docs/GettingStarted.md
```diff
@@ -115,15 +115,15 @@ Next:
       end
     end
 
-Is the implementation of our echo server. We define a class that inherits from {EventMachine::Connection}
+is the implementation of our echo server. We define a class that inherits from {EventMachine::Connection}
```
sodabrew, line 118:
```suggestion
are the implementation of our echo server. We define a class that inherits from {EventMachine::Connection}
```

ono-max:
@sodabrew 

I've updated it. Could you check this one again? 

Thanks!

PR #956, opened 2021-10-28
Add preverify_ok and X509_STORE_CTX to verify_peer

nevans:
This isn't complete by itself, because EM hasn't properly configured an
`X509_STORE` to go with its `SSL_CTX` and `SSL`, nor does it verify the
peer's identity (`verify_hostname` in stdlib).  Unlike stdlib, the
"store_context" is merely a read-only data structure, containing a copy
of the most useful data from the `X509_STORE_CTX`.

Also, the default implementation for `EM::Connection#ssl_verify_peer`
has been updated to log errors and returns `preverify_ok` unchanged.

### DIFF
--- a/ext/rubymain.cpp
+++ b/ext/rubymain.cpp
@@ -57,6 +57,7 @@ Statics
 
 static VALUE EmModule;
 static VALUE EmConnection;
+
 static VALUE EmConnsHash;
 static VALUE EmTimersHash;
 
@@ -78,7 +79,6 @@ static VALUE Intern_call;
 static VALUE Intern_at;
 static VALUE Intern_receive_data;
 static VALUE Intern_ssl_handshake_completed;
-static VALUE Intern_ssl_verify_peer;
 static VALUE Intern_notify_readable;
 static VALUE Intern_notify_writable;
 static VALUE Intern_proxy_target_unbound;
@@ -110,7 +110,6 @@ static inline VALUE ensure_conn(const uintptr_t signature)
 		rb_raise (EM_eConnectionNotBound, "unknown connection: %" PRIFBSIG, signature);
 	return conn;
 }
-
 
 /****************
 t_event_callback
@@ -188,9 +187,10 @@ static inline VALUE event_callback (VALUE e_value)
 		case EM_SSL_VERIFY:
 		{
 			VALUE conn = ensure_conn(signature);
-			VALUE should_accept = rb_funcall (conn, Intern_ssl_verify_peer, 1, rb_str_new(data_str, data_num));
-			if (RTEST(should_accept))
+			X509_STORE_CTX *ctx = (X509_STORE_CTX *)data_str;
+			if (em_ssl_verify_cb_call(conn, data_num, ctx))
 				evma_accept_ssl_peer (signature);
+
 			return Qnil;
 		}
 		#endif
@@ -1479,7 +1479,6 @@ extern "C" void Init_rubyeventmachine()
 	Intern_at = rb_intern("at");
 	Intern_receive_data = rb_intern ("receive_data");
 	Intern_ssl_handshake_completed = rb_intern ("ssl_handshake_completed");
-	Intern_ssl_verify_peer = rb_intern ("ssl_verify_peer");
 	Intern_notify_readable = rb_intern ("notify_readable");
 	Intern_notify_writable = rb_intern ("notify_writable");
 	Intern_proxy_target_unbound = rb_intern ("proxy_target_unbound");
@@ -1611,6 +1610,8 @@ extern "C" void Init_rubyeventmachine()
 	rb_define_const (EmModule, "SslVerify",                INT2NUM(EM_SSL_VERIFY                ));
 	// EM_PROXY_TARGET_UNBOUND = 110,
 	// EM_PROXY_COMPLETED = 111
+
+	Init_em_ssl();
 
 	// SSL Protocols
 	rb_define_const (EmModule, "EM_PROTO_SSLv2",   INT2NUM(EM_PROTO_SSLv2  ));
--- a/lib/eventmachine.rb
+++ b/lib/eventmachine.rb
@@ -24,6 +24,7 @@ require 'em/buftok'
 require 'em/timers'
 require 'em/protocols'
 require 'em/connection'
+require 'em/ssl'
 require 'em/callback'
 require 'em/queue'
 require 'em/channel'
@@ -1538,7 +1539,9 @@ module EventMachine
       c.ssl_handshake_completed
     elsif opcode == SslVerify
       c = @conns[conn_binding] or raise ConnectionNotBound, "received SslVerify for unknown signature: #{conn_binding}"
-      result = c.ssl_verify_peer(data) or c.close_connection
+      result = (c.original_method(:ssl_verify_peer).arity == 1) ?
+        c.ssl_verify_peer(data[1].current_cert.to_pem) : c.ssl_verify_peer(*data)
+      c.close_connection unless result
       result
     elsif opcode == TimerFired
       t = @timers.delete( data )
--- a/BSDL-openssl
+++ b/BSDL-openssl
@@ -1,0 +1,22 @@
+Copyright (C) 1993-2013 Yukihiro Matsumoto. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+SUCH DAMAGE.
--- a/COPYING-openssl
+++ b/COPYING-openssl
@@ -1,0 +1,56 @@
+Ruby is copyrighted free software by Yukihiro Matsumoto <matz@netlab.jp>.
+You can redistribute it and/or modify it under either the terms of the
+2-clause BSDL (see the file BSDL-openssl), or the conditions below:
+
+  1. You may make and give away verbatim copies of the source form of the
+     software without restriction, provided that you duplicate all of the
+     original copyright notices and associated disclaimers.
+
+  2. You may modify your copy of the software in any way, provided that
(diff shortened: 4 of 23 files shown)


nevans:
@MSP-Greg yeah, you're right.  I hadn't checked to see how far back EM is maintaining support!  `&.` is just a little bit of sugar, not worth upgrading over.  I'll push an update and run tests in different rubies locally.

Although the last time tried, I wasn't even able to compile some of the older versions, specifically because of OpenSSL library versions (IIRC).  I see the gemspec is set to 2.0.0, maybe that should be updated to 2.2 to match CI?

And I just now looked into the gh action repo to see if I could just grab one of those images and put it into docker, and I saw this:

> Note that Ruby ≤ 2.4 and the OpenSSL version it needs (1.0.2) are both end-of-life,
> which means Ruby ≤ 2.4 is unmaintained and considered insecure.

-- [ruby/setup-ruby README.md](https://github.com/ruby/setup-ruby/blob/b00551441a6fa3d528f3443dbff264a934b6e316/README.md?plain=1#L33-L34)

IMO, _that_ might be a good reason to drop support for older rubies! 😉

MSP-Greg:
> the gemspec is set to 2.0.0, maybe that should be updated to 2.2 to match CI?

The main people involved in `setup-ruby` are also involved with Ruby master, so there might be a bias there.  The Actions code here was committed before Ruby 2.1 and 2.0 were available in Actions (2.1 was added Jul 16, 2020, 2.0 was later).

I'm not the best person to ask about 'old Ruby support', but if it currently works, I've always been inclined to leave it included.

I'll see if it works in my fork...  I'm aware of the OpenSSL EOL issue, and I do wonder why people don't want to/can't update their apps.  I suppose it does support TLS 1.2, sort of...

nevans:
I'd like to make some other changes to the API, to give access to error (Integer), error_string, depth, _maybe_ even `error = code`.  In short: a wrapper for the X509_STORE_CTX.  ;)

IMO, it's a real pain in the neck to debug issues if you don't have all of that.  I'll update this PR hopefully some time later this week or weekend with my proposed API.

In the meantime, I have an bigger experimental branch that I'm interested in feedback on. I'll post a draft PR later today (I want to do a few rebase cleanups to make it easier to read first)

nevans:
Okay, I'll update this PR to work with different versions of ruby, but please also see #957

nevans:
It's passing in 2.4 - 3.0 now, but 2.0-2.3 don't pass in master (for me).  I rebased it on top of #958 and #959.

nevans:
FYI: that last commit updated the API to be closer to how I'd prefer it.  What do you think? (still waiting for the various tests to run).

nevans:
I haven't added tests for the new StoreContext object, yet. I'll update the existing ssl_verify tests to look at that too.

nevans:
haha, I swear I'm not pushing until it for me locally.  I suppose I shouldn't be surprised that the behavior here is slightly different, but I am a bit puzzled.  I'm running `OpenSSL 1.1.1l  24 Aug 2021` on Ubuntu 21.10, using ruby versions via `rbenv install`...

I think the behavior that's different is that, when you override errors, _sometimes_ openssl calls the verify callback once per error and then a final time per certificate with no error, and other times it only calls once per error.  These test failures are all displaying "once per error, w/o a non-error" behavior, but locally I get that final non-error (as shown in the tests).

If you know anywhere that it's documented why this happens, or some way to regularize the behavior, that'd be great.  Otherwise, I'll just relax the tests to not care about that final non-error callback.

nevans:
FYI: this is passing all builds for me on my branch now.

ext/rubymain.cpp
```diff
@@ -111,6 +115,155 @@ static inline VALUE ensure_conn(const uintptr_t signature)
...
+
+#ifdef WITH_SSL
+
+// adapted from stdlib openssl's ossl_str_new_i
+static VALUE em_ssl_str_new_i(VALUE size)
+{
+	return rb_str_new(NULL, (long)size);
+}
+
+// adapted from stdlib openssl's ossl_str_new
```
sodabrew, line 130: Would it be possible to use those method directly (if they are part of the public API?)

nevans: For the most part, yes, and I would love to just use the methods directly.  Even better, I'd love to just use `OpenSSL::X509::StoreContext` directly, and skip all of the re-implementation.

The problem (as I understand it) is that there's no mechanism for a ruby gem to publish C header files that can safely be used (not even for bundled or default gems, like `openssl`).  And even then, we'd also need the `openssl` gem to commit to a public API for any C funcs that we used... many of the functions I want access to are static and not exported.  There's at least one b.r-l.org conversation about this issue: https://bugs.ruby-lang.org/issues/17760, but it didn't really come to any resolution, AFAICT.

That said, it _should_ be possible to find the gem's install dir and use ext/openssl/ossl.h.  And the most important functions that we might want to access _are_ exported.  I just don't have any idea how brittle that would be... and _when_ it breaks, how hard to debug will it be?

sodabrew: What you've done makes sense given that these are internal methods that aren't being _intentionally_ exported with a stable API. Someone put the idea in my head ages ago to move all of the TLS work up into Ruby space and implement it using the ruby/openssl standard library gem, and just rip out all of the C++ TLS bits. That's not something I've had any time in years to think much about, but would seem to be the right move.

nevans: Adding a few more links I found:
* https://bugs.ruby-lang.org/issues/17760
* https://bugs.ruby-lang.org/issues/17761 (Fixed by: https://github.com/ruby/ruby/commit/5cdf99f6)
* https://bugs.ruby-lang.org/issues/15156
* https://github.com/rubygems/rubygems/issues/2408
  * https://github.com/rubygems/rubygems/pull/2416
* https://github.com/ruby/openssl/issues/475
  _I created this issue on the openssl gem back when I first worked on this current eventmachine PR._

sodabrew:
Let me know if you're still working on this PR? I think it looks good

lib/eventmachine.rb
```diff
@@ -1538,7 +1539,9 @@ def self.event_callback conn_binding, opcode, data
       c.ssl_handshake_completed
     elsif opcode == SslVerify
       c = @conns[conn_binding] or raise ConnectionNotBound, "received SslVerify for unknown signature: #{conn_binding}"
-      result = c.ssl_verify_peer(data) or c.close_connection
+      result = (c.original_method(:ssl_verify_peer).arity == 1) ?
```
sodabrew, line 1542: Suggest to add a comment here that arity == 1 is for backwards compatibility

nevans:
> Let me know if you're still working on this PR? I think it looks good

Thanks!  Not quite yet.  I should have it ready some time tomorrow.

nevans:
For what it's worth, I _did_ find quite a few examples of gems that link against other gem extensions:
* The [`ruby-gnome`](https://github.com/ruby-gnome/ruby-gnome) project actually seems to have a lot of infrastructure based around this:
  * [`mkmf-gnome`](https://github.com/ruby-gnome/ruby-gnome/blob/e10cbfde4f6e984766bf93323532a08c416c2dd8/glib2/lib/mkmf-gnome.rb#L461) has several methods which support linking against other gems
  * See e.g: [`add_depend_package(gem_name, ...)`](https://github.com/ruby-gnome/ruby-gnome/blob/e10cbfde4f6e984766bf93323532a08c416c2dd8/glib2/lib/mkmf-gnome.rb#L156-L192), which is used by [17 different gems in the `ruby-gnome` repo](https://github.com/search?q=repo%3Aruby-gnome%2Fruby-gnome+add_depend_package+path%3A**%2Fextconf.rb&type=code), and [many other gems outside it](https://github.com/search?q=add_depend_package+path%3A**%2Fextconf.rb+NOT+is%3Afork&type=code)
* I found many gems linking against [`narray`](https://github.com/masa16/narray), its successor [`numo-narray`](https://github.com/ruby-numo/numo-narray), and [`nmatrix`](https://github.com/SciRuby/nmatrix).  For example:
  * [`cumo`](https://github.com/sonots/cumo/blob/d59d507cb03835a85d3d784d55e4f28125729dab/ext/cumo/extconf.rb#L13-L29)
  * [`gsl`](https://github.com/SciRuby/rb-gsl/blob/103a3e1941e220befd037e0f4289a82c551ae250/ext/gsl_native/extconf.rb#L59-L84)
  * [`cuda-gaussian1d`](https://github.com/Himeyama/cuda-gaussian1d/blob/3a8bb04d81d9ad798fa681dbb9831265490b4a71/ext/cuda/gaussian1d/extconf.rb#L5-L16)
  * [`cuda-convolve`](https://github.com/Himeyama/cuda-convolve/blob/7c2aea0ec46392a787dd83e1596c631401a520a7/ext/cuda/convolve/extconf.rb#L5-L17)
  * [`arrayfire`](https://github.com/arrayfire/arrayfire-rb/blob/21179a1bd2d351772eb510951d433e84735e74d9/ext/mri/extconf.rb#L5-L49)
  * [`ruby-mpi`](https://github.com/gfd-dennou-club/ruby-mpi/blob/fed22f98386b65d1f12ca014ecc9744d59d1d08b/ext/mpi/extconf.rb#L7-L23)
* Linked against the `mysql2` gem:
  * [`mysql2-replication`](https://github.com/groonga/mysql2-replication/blob/c5c7565d6ecc43e45ecf73193c0fc18b00209dcd/ext/mysql2-replication/extconf.rb#L46-L48)
  * [`mysql2-arrow`](https://github.com/red-data-tools/activerecord-arrow-adapters/blob/b855c7698eb8a1e18c6c0dbf0a3adb120039b101/mysql2-arrow/ext/mysql2_arrow/extconf.rb#L61-L65)
* [`libxslt-ruby`](https://github.com/xml4r/libxslt-ruby/blob/253b00a0b52d277d5bb559a692fadcd0c2dced4a/ext/libxslt/extconf.rb#L75-L83) links to `libxml-ruby`
* [`debase-ruby_core_source`](https://github.com/ruby-debug/debase-ruby_core_source) looks like it primarily exists so other gems can link against ruby's internal headers:
  * [`debase`](https://github.com/ruby-debug/debase/blob/551b63d4f53ba6ded8f753725baa16859f673751/ext/extconf.rb#L28-L34)
  * [`arg_scanner`](https://github.com/JetBrains/ruby-type-inference/blob/df63525a226c4926614a3937546b570b68bc42aa/arg_scanner/ext/arg_scanner/extconf.rb#L44-L52)
  * [`perftools.rb`](https://github.com/tmm1/perftools.rb/blob/42ba17b770a7c9d9c92bce582eca7b025b91de62/ext/extconf.rb#L14-L42)
  * [`datadog` (aka `ddtrace`)](https://github.com/DataDog/dd-trace-rb/blob/c5ab063df619ea6557a9124871d56d2fc72f6999/ext/datadog_profiling_native_extension/extconf.rb#L272-L292)
* `xnd-ruby` gems:
  * [`gumath`](https://github.com/xnd-project/xnd-ruby/blob/ddecb88d3a6fa17ca314b6599aedf4f5a189b067/gumath/ext/ruby_gumath/extconf.rb#L17-L45) links against `xnd` and `ndtypes`
  * [`xnd`](https://github.com/xnd-project/xnd-ruby/blob/ddecb88d3a6fa17ca314b6599aedf4f5a189b067/xnd/ext/ruby_xnd/extconf.rb#L17-L29) links against `ndtypes`
* [`nokogumbo`](https://github.com/rubys/nokogumbo/blob/982f9676887057efad5b47cd160f015fb1a3fa2c/ext/nokogumbo/extconf.rb#L11) linked against the `nokogiri` gem
  _(nokogumbo was merged into nokogiri several years ago, but it's still a useful example of the pattern)_
* [`rugged-redis`](https://github.com/redbadger/rugged-redis/blob/19a9168e7c423238f5990c97bc98713246043684/ext/rugged/redis/extconf.rb#L31-L42) links against the `rugged` and `hiredis` gems
  _(not updated in 7 years, but still a useful example of the pattern)_
* [`rugged-mysql`](https://github.com/chonglou/rugged-mysql/blob/d9d64201f7ba84a1281f4b8648deef392ec6165a/ext/rugged/mysql/extconf.rb#L7-L11) links against the `rugged` gem
  _(not updated in 10 years, but still a useful example of the pattern)_
* [`ruby-libappindicator`](https://github.com/rubiojr/ruby-libappindicator/blob/b5b4a8b66a5c00f75629eabac60b383291e9dbdb/ext/libappindicator/extconf.rb#L15-L34) links against the `glib2` and `gtk2` gems
  _(not updated in 13 years, but still a useful example of the pattern)_

But I have questions and concerns about stability.  Is this a safe thing to do?  Under what circumstances can they break?  Some of these gems have millions of downloads, so the pattern does seem to have _some_ resilience.

Is there a way for a gem to signal that all of the exported methods in a particular header file will be ABI backward compatible within a version range (besides just asking in their discussion forums or issue trackers or adding it to their README or other docs)?

nevans:
@sodabrew Those examples do give me some confidence that we can make that approach work.  That said, I'll finish up this PR in its current form (with the copied methods).  I'll make another PR that explores linking directly against stdlib `openssl`.

PR #958, opened 2021-11-13
Generate test CA and certificates in rake

nevans:
As part of tidying up the SSL implementation, I'm updating all of the
tests to use secure defaults.  This commit doesn't enable certificate
verification or identity verification, yet.  This just adds a little bit
of infrastructure for quickly testing several different SSL scenarios.

Currently, it's just a single CA which has signed a single "localhost"
certificate.  I'll update it to use intermediate certs and multiple
hostnames, etc, in future PRs.


nevans:
Passed tests for me in C ruby 2.4.10, 2.5.9, s.6.8, 2.7.4, 3.0.2.
I haven't tested 2.0, 2.1, 2.2, or 2.3.  I'll need to do that via docker images.
Failing in jruby and truffleruby, but so is master. ;)

nevans:
FYI: I didn't get master to compile in docker for 2.2 or 2.3.  Perhaps the github actions set things up better than the official docker images?

MSP-Greg:
Maybe run GitHub Actions in your fork?

nevans:
Ah, thanks. Took me a minute to figure out how to run actions from my phone 🙃 And I see you have your own Ruby setup action, which makes sense, since nice if the others seem to work with old ruby & old openssl.

Anyway, they all passed except for a MacOS 2.6 🤔https://github.com/nevans/eventmachine/runs/4198820287?check_suite_focus=true

MSP-Greg:
Glad you got Actions running.  Not sure about the macOS issue...

> you have your own Ruby setup action

MSP-Greg/setup-ruby-pkgs uses the code from ruby/setup-ruby to install Ruby, it just adds the ability to install packages, in this case, openssl on Windows.  It also works with apt and brew.

I've helped several repos convert to Actions, as it runs a high number of concurrent jobs, often speeding up CI...

nevans:
Well, they're all passing now! Flaky test? Could the private keys be generated in such a way that key[100] is insignificant? If so, perhaps something like the following could be less flaky?

````ruby
idx = badkey.length / 2
badkey[idx] = badkey[idx].tr("A-Za-z", "N-ZA-Mn-za-m")
````

nevans:
(that specific `tr` wouldn't work if the characters aren't alphabetic, but that's simple to fix with a second `tr`)

nevans:
Any thoughts on this (two and a half year old) PR?  It's purely a refactoring of the tests.  I can simplify it, if you'd like.

PR #959, opened 2021-11-13
Add .editorconfig for rb, cpp, h files

nevans:
I was just using editorconfig settings that seemed sensible for what's already there.  Please change the actual values to whatever seems most suitable to you.  😄 

I suppose I *could* add "compile_commands.json" to my global gitignore... But IMO adding it here is reasonable.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+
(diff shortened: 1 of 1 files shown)


nevans:
Also, would you mind if I add a .clang-format?  I'll try to make it fit the existing formatting as best I can, but... I'll unavoidably wind up inserting some of my own preferences. 😉  I can do that in a separate PR.

PR #961, opened 2021-11-22
Increase timeout on a flaky test

nevans:
In my local development, this was the only truly flaky test.  It's not a
performance test and it's reliant on an external resource, so IMO adding
an extra second to its run time should be fine.


nevans:
Any thoughts on this (two and a half year old) PR? It's purely a refactoring of a single flaky test. I could update it to use an environment variable, if you'd like.

nevans:
What's in #995 seems to work better than this (and covers many other timeouts, too).  That PR sets `TIMEOUT = TIMEOUT_INTERVAL * 4.0`, which automatically increases _all_ of the uses of `TIMEOUT` in this file.

PR #967, opened 2022-04-14
there are bugs in if-statements using `and`.

mephistobooks:
the operators of `and`, `or`, and `not` in if-statements are too buggy.
(ex. `p true and false` =>true)
we should not use them. actually there are bugs  in `eventmachine.rb` so i correct them.

please read this: "Ruby Style Guide", [no `and`/`or`](https://github.com/rubocop/ruby-style-guide#no-and-or-or).


sodabrew:
Is there an example of a situation where one of these resulted in a wrong output?

mephistobooks:
For example,

```
if defined?(EventMachine.library_type) and EventMachine.library_type == :pure_ruby
```

this is always true even if `EventMachine.library_type` is just defined. i.e., it becomes true also when `EventMachine.library_type != :pure_ruby`... I would think this is not expected behavior.

sodabrew:
That would be a surprising result, but I'm not able to reproduce that in a simple `irb` session. What am I missing?

mephistobooks:
Oh, really...? Simply we can check it as follows:
```
x = :pure_ruby
p true and x == :pure_ruby   # this becomes true. yeah, it's natural.
# =>true
p true and x != :pure_ruby    # but this becomes also true... 😭
# =>true
```
Both we get are true. So we must say that the codes using `and`/`or` in `if`-conditions have semantically (or potentially) bug. It is VERY dangerous...

MSP-Greg:
@mephistobooks

First, I would agree that using `and`/`or` can have unintended consequences.

Your example is incorrect, mostly because `p`/`puts` are method calls, and `if` is not.  It also shows how problems can occur when not using parentheses with method calls.  Try:
```ruby
x = :pure_ruby
if true and x == :pure_ruby
  p 'true'
else
  p 'false'
end

if true and x != :pure_ruby
  p 'true'
else
  p 'false'
end
```

sodabrew:
Thanks @MSP-Greg, exactly the explanation for this situation of using `puts` and getting a different result.

@mephistobooks I would be glad to take the PR for modern Ruby style if you would clarify the commit message to say that this is a style change not a bug-fix.

mephistobooks:
@sodabrew I'm sorry, @MSP-Greg is correct. I'm wrong (I misunderstood it with assignment expression).

nevans:
FWIW, although I agree with every change in this PR, I resent the ruby style guide total ban.  And the justification confuses more than it clarifies (basically: it's FUD).  So I'm just registering that complaint.  😜  Avdi Grimm has [a good explanation](http://www.virtuouscode.com/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/) (transcript available at the unlinkable "Video Script" tab).  But, Avdi and I both learned Perl before ruby.

Fortunately, the rubocop default is [`Style/AndOr: {EnforcedStyle: conditionals}`](https://docs.rubocop.org/rubocop/cops_style.html#enforcedstyle-conditionals-default), which (mostly) agrees with Avdi and me, and not with the style guide. 😉

Contrary to what the style guide says: ruby's precedence makes more sense than Perl's (IMO) once you understand the design rationale. 🙂  It makes perfect sense for boolean algebra symbols to follow the standard math precedence rules (disjunctive normal form, or the sum of products) and the English control flow words to follow the standard linguistic rules (simple left-to-right).

Anyway, everyone's time might've been better served if, instead of writing this comment, I had submitted PRs to the ruby style guide and ruby's `doc/syntax/{precedence,control_expressions}.rdoc`!

PR #974, opened 2022-09-13
Fix TLS connection of the pure Ruby reactor

meineerde:
With these commits, I was able to successfully run `rake test_pure` without any failing tests.

### DIFF
--- a/lib/em/pure_ruby.rb
+++ b/lib/em/pure_ruby.rb
@@ -280,7 +280,7 @@ module EventMachine
       @tls_parms[signature][:priv_key] = priv_key if tls_parm_set?(priv_key)
       @tls_parms[signature][:priv_key_pass] = priv_key_pass if tls_parm_set?(priv_key_pass)
       @tls_parms[signature][:cert_chain] = File.read(cert_chain_path) if tls_parm_set?(cert_chain_path)
-      @tls_parms[signature][:cert_chain] = cert if tls_parm_set?(cert_chain)
+      @tls_parms[signature][:cert_chain] = cert if tls_parm_set?(cert)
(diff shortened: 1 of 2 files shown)


sodabrew:
Thank you! Running CI now

sodabrew:
Actually, it occurs to me that the GitHub CI test suite does not have a pure-ruby mode. Would you be up for adding it at least one row to the matrix with one version of Ruby that runs it?

meineerde:
I added a test step. Unfortunately, some tests do still fail (on my Mac) when running `rake test_em_pure_ruby`. My previous changes improve the situation compared to what was before, but apparently there is still some differing behavior...

It might still be necessary to mark the build test as not required (i.e. something like `allow_failures` on Travis CI), but I'm not 100% sure how to correctly configure this with Github Actions.

lib/em/pure_ruby.rb
```diff
@@ -280,7 +280,7 @@ def set_tls_parms signature, priv_key_path, priv_key, priv_key_pass, cert_chain_
       @tls_parms[signature][:priv_key] = priv_key if tls_parm_set?(priv_key)
       @tls_parms[signature][:priv_key_pass] = priv_key_pass if tls_parm_set?(priv_key_pass)
       @tls_parms[signature][:cert_chain] = File.read(cert_chain_path) if tls_parm_set?(cert_chain_path)
-      @tls_parms[signature][:cert_chain] = cert if tls_parm_set?(cert_chain)
+      @tls_parms[signature][:cert_chain] = cert if tls_parm_set?(cert)
```
sodabrew, line 283: I'm looking if this was a bug previously? cert vs. cert_chain in the if block

sodabrew:
I posted a question here, and otherwise this PR looks subsumed into #993 cc @nevans -- Thank you both!!

PR #979, opened 2023-04-05
Set TLS params: Don't use deprecated functions anymore

marek22k:


sodabrew:
lgtm

lib/em/pure_ruby.rb
```diff
@@ -316,22 +316,18 @@ def start_tls signature
       end
       ctx.ciphers = tls_parms[:cipher_list] if tls_parms[:cipher_list]
       if selectable.is_server
-        ctx.tmp_dh_callback = Proc.new do |_, _, key_length|
-          if tls_parms[:dhparam]
-            OpenSSL::PKey::DH.new(tls_parms[:dhparam])
+        ctx.tmp_dh = if tls_parms[:dhparam]
+          OpenSSL::PKey::DH.new(tls_parms[:dhparam])
+        else
+          case key_length
```
sodabrew, line 322: Oops, didn't notice that `key_length` is now undefined. It was previously an argument to the `tmp_dh_callback` method.

nevans: @sodabrew I handled that in #993, after I rebased on top of this PR.  :)

PR #988, opened 2024-07-22
Fix usage of Process::Status on ruby3.3

terceiro:


ext/rubymain.cpp
```diff
@@ -563,6 +563,13 @@ static VALUE t_get_subprocess_status (VALUE self UNUSED, VALUE signature)
 	if (evma_get_subprocess_status (NUM2BSIG (signature), &status)) {
 		if (evma_get_subprocess_pid (NUM2BSIG (signature), &pid)) {
 
+#ifdef IS_RUBY_3_3_OR_LATER
```
sodabrew, line 566:
```suggestion
#if defined(IS_RUBY_3_3_OR_LATER)
```

ext/rubymain.cpp
```diff
@@ -563,6 +563,13 @@ static VALUE t_get_subprocess_status (VALUE self UNUSED, VALUE signature)
...
+			proc_status = rb_obj_alloc(rb_cProcessStatus);
+			struct rb_process_status *data = NULL;
+			data = (rb_process_status*)RTYPEDDATA_GET_DATA(proc_status);
+			data->pid = pid;
+			data->status = status;
+#else
 #ifdef IS_RUBY_3_OR_LATER
```
sodabrew, line 573:
```suggestion
#elif defined(IS_RUBY_3_OR_LATER)
```

ext/rubymain.cpp
```diff
@@ -585,6 +592,7 @@ static VALUE t_get_subprocess_status (VALUE self UNUSED, VALUE signature)
 			rb_ivar_set(proc_status, rb_intern_const("status"), INT2FIX(status));
 			rb_ivar_set(proc_status, rb_intern_const("pid"), INT2FIX(pid));
 #endif
+#endif
```
sodabrew, line 595:
```suggestion
```

sodabrew:
Thank you! I think this will be a bit clearer with `#if defined()`

terceiro:
@sodabrew updated, thanks for the review

PR #993, opened 2024-08-30
Fix pure ruby tests

nevans:
This is a collection of fixes for making the test suite pass using the pure ruby implementation.  I started by splitting these each into their own atomic PRs, but there was quite a bit to do than I originally realized.  So I wanted to have a single PR that fixes all of the tests _and_ enables the pure ruby workflow in GitHub Actions.

Please note: this PR includes the following others:
* #991
* #992 

And, although this gets the pure ruby CI passing, the regular CI suite is not currently passing.  This PR needs to be merged first before the full build can pass.
* #995

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -79,7 +79,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, ubuntu-24.04 ]
+        os: [ ubuntu-20.04, ubuntu-22.04, ubuntu-24.04 ]
         ruby: [ '2.7', '3.0', '3.1', '3.2', '3.3', head ]
     steps:
       - name: repo checkout
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -71,8 +71,41 @@ jobs:
           CI: true
           TESTOPTS: -v --no-show-detail-immediately
 
+  pure_ruby:
+    name: >-
+      pure ruby (${{ matrix.os }} ${{ matrix.ruby }})
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-22.04, ubuntu-24.04, macos-14, windows-2022 ]
+        ruby: [ '2.7', '3.0', '3.1', '3.2', '3.3', head ]
+        include:
+          - { os: windows-2022, ruby: mingw }
+        exclude:
+          - { os: windows-2022, ruby: head }
+    steps:
+      - name: repo checkout
+        uses: actions/checkout@v4
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
+
+      - name: macOS disable firewall
+        if: startsWith(matrix.os, 'macos')
+        run: |
+          sudo /usr/libexec/ApplicationFirewall/socketfilterfw --setglobalstate off
+          sudo /usr/libexec/ApplicationFirewall/socketfilterfw --getglobalstate
+
+      - name: bundle install
+        run:  bundle install --jobs 4 --retry 3 --without=documentation
+
       - name: test_em_pure_ruby
         run:  bundle exec rake test_em_pure_ruby
         env:
           CI: true
+          EM_PURE_RUBY: true
           TESTOPTS: -v --no-show-detail-immediately
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -79,7 +79,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, ubuntu-24.04, macos-14, windows-2022 ]
+        os: [ ubuntu-22.04, ubuntu-24.04, macos-latest, windows-2022 ]
         ruby: [ '2.7', '3.0', '3.1', '3.2', '3.3', head ]
         include:
           - { os: windows-2022, ruby: mingw }
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -79,12 +79,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, ubuntu-24.04, macos-latest, windows-2022 ]
+        os: [ ubuntu-22.04, ubuntu-24.04 ]
         ruby: [ '2.7', '3.0', '3.1', '3.2', '3.3', head ]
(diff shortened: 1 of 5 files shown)


sodabrew:
Phenomenal! Approving test suite on GitHub now

Gemfile
```diff
@@ -8,6 +8,11 @@ install_if -> { RUBY_VERSION > '3.1' } do
   gem "net-smtp"
 end
 
+# switch to install_if when ruby 2.2 support is dropped
```
sodabrew, line 11: Please do switch to install_if now, there's one just above this block now

lib/em/pure_ruby.rb
```diff
@@ -927,32 +1005,71 @@ def self.connect bind_addr, bind_port, host, port
...
+    rescue OpenSSL::SSL::SSLError => error
+      warn "SSL Error in EventMachine check_handshake_complete: #{error}"
+      @ssl_handshake_state = error
+    rescue => error
+      warn "#{error.class} in EventMachine check_handshake_complete: #{error}"
+      @ssl_handshake_state = error
     end
 
     def pending?
-      handshake_complete?
```
sodabrew, line 955: Seems likely this method was internal use only, but calling it out to think/confirm

lib/em/pure_ruby.rb
```diff
@@ -927,32 +1005,71 @@ def self.connect bind_addr, bind_port, host, port
...
-      @handshake_complete = false
+      @ssl_handshake_state = nil
     end
 
+    TCP_ESTABLISHED = 1 # why isn't this already a const in Socket?
+
     def ready?
-      if RUBY_PLATFORM =~ /linux/
-        io.getsockopt(Socket::SOL_TCP, Socket::TCP_INFO).unpack("i").first == 1 # TCP_ESTABLISHED
+      if defined?(Socket::SOL_TCP) && defined?(Socket::TCP_INFO)
```
sodabrew, line 1014: Nice

lib/em/pure_ruby.rb
```diff
@@ -265,51 +288,95 @@ def tls_parm_set?(parm)
...
         :ssl_options => ssl_options
       }
-      @tls_parms[signature][:priv_key] = File.read(priv_key_path) if tls_parm_set?(priv_key_path)
+      @tls_parms[signature][:priv_key] = File.binread(priv_key_path) if tls_parm_set?(priv_key_path)
       @tls_parms[signature][:priv_key] = priv_key if tls_parm_set?(priv_key)
       @tls_parms[signature][:priv_key_pass] = priv_key_pass if tls_parm_set?(priv_key_pass)
-      @tls_parms[signature][:cert_chain] = File.read(cert_chain_path) if tls_parm_set?(cert_chain_path)
-      @tls_parms[signature][:cert_chain] = cert if tls_parm_set?(cert_chain)
+      @tls_parms[signature][:cert_chain] = File.binread(cert_chain_path) if tls_parm_set?(cert_chain_path)
+      @tls_parms[signature][:cert_chain] = cert if tls_parm_set?(cert)
```
sodabrew, line 325: Should this stay `... if tls_parm_set?(cert_chain)` or was it incorrect previously?

lib/em/pure_ruby.rb
nevans: Yeah, unfortunately, my efforts didn't result in a fully working version for windows and macos.  I'd say that there's still some refactoring to do in that area of the pure ruby reactor to bring it to parity with the cpp version on all platforms.

lib/em/pure_ruby.rb
nevans: Yeah, `#handshake_complete?` seems to have only existed in the pure_ruby selector... it seemed similar in function to `ConnectionDescriptor::_CheckHandshakeStatus` in the cpp.  I renamed it to (or replaced it with) `#check_handshake_complete`, because calling predicate methods for their side effects confuses me (also it matches the name of its related cpp method a little closer now).

nevans:
> I merged #993 for contribution attribution purposes, hopefully will be an easy enough rebase to sit your changes atop those. The 4096 key length for example, is a good change to go beyond the 2048 of that PR.

Great.  I was considering rebasing on top of that PR for attribution purposes anyway.  :)

sodabrew:
```
NameError: undefined local variable or method `key_length' for EventMachine:Module
```

nevans:
> ```
> NameError: undefined local variable or method `key_length' for EventMachine:Module
> ```
> 
> Came in from [ecba994](https://github.com/eventmachine/eventmachine/commit/ecba994cd55ee21a74fc215eb876f38465265e6f)

Rather than bring back the callback (which could react to `key_length), I chose to fix it by copying stdlib's OpenSSL::SSL::SSLContext::DH_ffdhe2048, which is copied from RFC7919: 37200483.

nevans:
I _did_ check to see if `tmp_dh=` is defined, since it was only added in openssl gem 3.1+ (shipped with ruby 3.2, compatible with ruby 2.6+).

Gemfile
nevans: Done, and thanks for dropping 2.2 from the build.  I was surprised the bundler that came with ruby 2.2 couldn't handle the `install_if` version, but I didn't want to resort to multiple gemfiles!

.github/workflows/workflow.yml
```diff
@@ -71,8 +71,31 @@ jobs:
...
+  pure_ruby:
+    name: >-
+      pure ruby (${{ matrix.os }} ${{ matrix.ruby }})
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-20.04, ubuntu-22.04, ubuntu-24.04 ]
+        ruby: [ '2.7', '3.0', '3.1', '3.2', '3.3', head ]
```
nevans, line 83: @sodabrew What do you think about only running pure ruby CI on non-EOL ruby versions?
```suggestion
        ruby: [ '3.1', '3.2', '3.3', head ]
```

nevans:
@sodabrew the only current build failures are the flaky ones.  I might take another stab at making those less flaky later (i.e: not today).

.github/workflows/workflow.yml
sodabrew: Oh 2.7 is deprecated already, is it 😭

Gemfile
```diff
@@ -8,6 +8,11 @@ install_if -> { RUBY_VERSION > '3.1' } do
   gem 'net-smtp'
 end
 
+# switch to install_if when ruby 2.2 support is dropped
+if RUBY_VERSION >= '3.0'
+  gem "sorted_set"
```
sodabrew, line 13:
```suggestion
  gem 'sorted_set'
```

nevans:
@sodabrew Thanks!

PR #995, opened 2024-09-03
Update CI matrix

nevans:
This PR updates the GitHub Actions CI workflow, and all of its jobs should pass now.

### Update the job matrix
* **Drop unsupported `ubuntu-18.04` runner from CI**
  GitHub Actions dropped support for ubuntu-18.04 on 2023-04-03:
  - https://github.com/actions/runner-images/issues/6002

  Although there are benefits to testing with older OS versions—especially given EventMachine's legacy status and the types of projects which are most likely to depend on it—there are also significant costs to maintaining CI for them once GitHub Actions has dropped support.
* **Add newer Ubuntu runners to CI**
  Each Ubuntu LTS comes with a different version of OpenSSL, Linux Kernel, etc.  And people are likely to deploy EventMachine applications to any of these versions.  So there is some benefit to testing every Ubuntu LTS release that is supported by GitHub Actions.
* **Exclude ubuntu-22.04 / ruby 2.2 from CI**
  This avoids the following error:

      Bundler 2 requires Ruby 2.3+, using Bundler 1 on Ruby <= 2.2
      /opt/hostedtoolcache/Ruby/2.2.10/x64/bin/gem install bundler -v ~> 1.0
      ERROR:  While executing gem ... (RuntimeError)
          Marshal.load reentered at marshal_load

  See https://github.com/ruby/setup-ruby/issues/496

  Ubuntu 20.04 and 24.04 do not have this issue.
* **Switch CI to use `macos-latest` runner**
  GitHub Actions dropped support for `macos-10.15` on 2022-12-01:
  - https://github.com/actions/runner-images/issues/5583

  Under the presumption that MacOS is primarily used for development and not for deployment, I've switched the macos GitHub Actions runners to use the "latest" release version for ruby >= 2.6.  Earlier ruby's don't support arm64, so they are using the macos-13 runner (which is amd64).

  On the one hand, this is less explicit and is no longer "pinned", so it may result in CI failing when GitHub Actions updates these runners.  On the other hand, we also won't need another PR to update these every time an old version is dropped or a new version is added.

  I attempted to make the same change from `windows-2019` to `windows-latest`.  But I saw three tests that reliably pass under windows-2019 and reliably fail under `windows-2022`.  Since I don't have a Windows OS dev env at my disposal, I'm opting to just let windows CI continue under `windows-2019`, for now.
* **Add newer ruby versions to the CI matrix**
  Ruby 3.2 should pass in CI.  However, ruby 3.3 (and head) will not pass until #988 (or something like it) is merged.  Windows mingw is likewise excluded; it builds ruby HEAD.

### Update the actions used by CI steps
* **Update `actions/checkout` to v4.**  `actions/checkout@v2` has been deprecated for some time now.
* **Update `ruby/setup-ruby-pkgs` to ruby organization.**  It looks like `ruby/setup-ruby-pkgs` contains all of the commits in `MSP-Greg/setup-ruby-pkgs`.  I'm assuming that the ruby repository is now the official version and that new updates will be pushed there.  So this switches to that repo.
### Increase various CI timeouts
* **Raise `TIMEOUT_INTERVAL` and other test timeouts for windows and macos.**  Several of these tests appeared to be flaky in CI, but only for macos or windows.  So I went through and raised many of them, for those OSes only.  In many cases, I changed the timeouts to be a multiple of TIMEOUT_INTERVAL, so that individual tests don't each need to check if they are running on `windows? || darwin?`.
* **Double test_periodic_timer timeout (in TestPure).**  This timeout was flaky in CI, so I doubled the allowed delta.
* **Increase CI job's `timeout-minutes` for windows and macos runners.**  Sometimes these jobs take 8 minutes just installing ruby and related packages, and then another 4 minutes to compile eventmachine run the test suite.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-20.04, ubuntu-18.04, macos-10.15, windows-2019 ]
+        os: [ ubuntu-20.04, macos-10.15, windows-2019 ]
         ruby: [ '2.2', '2.3', '2.4', '2.5', '2.6', '2.7', '3.0', '3.1', head ]
         include:
           - { os: windows-2019, ruby: mingw }
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -51,7 +51,7 @@ jobs:
 
     steps:
       - name: repo checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: load ruby, openssl
         uses: MSP-Greg/setup-ruby-pkgs@v1
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -7,7 +7,7 @@ jobs:
     name: >-
       ${{ matrix.os }} ${{ matrix.ruby }}
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 10
+    timeout-minutes: ${{ (startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'windows')) && 15 || 10 }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-20.04, macos-10.15, windows-2019 ]
+        os: [ ubuntu-24.04, ubuntu-22.04, ubuntu-20.04, macos-10.15, windows-2019 ]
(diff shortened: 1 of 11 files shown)


nevans:
@sodabrew I know you don't have much time for maintenance of eventmachine these days--and I can relate!  But this PR is purely tests and CI, with no changes to actual code.  So I'm hoping it should be low risk, and easier to "pull the trigger" on than other PRs?

If you have a moment, maybe you can merge it or provide feedback?

Also, are there any others who could share the maintenance burden?
(cc: @MSP-Greg since you also wrote a fair bit of eventmachine code since the last release.)

nevans:
FYI: see the passing CI results here: https://github.com/nevans/eventmachine/actions/runs/10688818614

sodabrew:
I’ll check if I have admin privileges to add maintainers, though I don’t recall having them

PR #998, opened 2024-09-06
Remove most EOL rubies from the build matrix

### LINKED ISSUE #997: Next release - remove support for old Rubies?
EM currently supports Ruby 2.0 with the following in the gemspec `s.required_ruby_version = '>= 2.0.0'`.

CI is testing starting with Ruby 2.3.

Is it time to set the minimum Ruby version to 2.5 or greater?

nevans:
Most deployments will be to linux servers, and EventMachine *is* still maintaining *some* support for old versions of ruby.  So this still keeps up to four EOL ruby builds for ubuntu, but _only_ for ruby versions that went EOL _after_ that version of ubuntu was first released.

This fixes #997, following the verbose proposal I made in my comment on that ticket. 🙂

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -13,10 +13,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-24.04, ubuntu-22.04, ubuntu-20.04, windows-2022, macos-14, macos-13 ]
+        os:
+          - ubuntu-20.04
+          - ubuntu-22.04
+          - ubuntu-24.04
+          - macos-13 # amd64
+          - macos-14 # arm64
+          - windows-2022
+
         # We currently support four EOL versions of ruby, but only on linux
         # versions that were released before that version of ruby went EOL.
         ruby: [ '2.5', '2.6', '2.7', '3.0', '3.1', '3.2', '3.3', head ]
+
         include:
           - { os: windows-2022, ruby: ucrt } # used instead of "head"
         exclude:
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -56,7 +56,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: load ruby, run bundle install
-        uses: ruby/setup-ruby-pkgs@v1
+        uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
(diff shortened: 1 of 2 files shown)


MSP-Greg:
One of us will have to rebase (see #1000).  We'll see what @sodabrew wants to do.  I can clean up #1000 if we up to Ruby 2.5...

MSP-Greg (approved):

nevans:
@MSP-Greg I rebased this on top of #1000.  We can and should easily merge your workflow improvements regardless of whether or not we decide to merge this PR.

MSP-Greg:
Since Ruby 2.3 & 2.4 are removed, we can change the workflow's line 46 from `uses: ruby/setup-ruby-pkgs@v1` to `uses: ruby/setup-ruby@v1`.

Not sure how fast everyone wants CI to run, but adding some more Ruby 2.x jobs might be helpful.

Also, 'excluding' rather than 'including' places similar platform/os jobs in one group in the job list, which can be helpful when one is totally frustrated with CI issues...

sodabrew:
Oh, I took these out of order. Let's keep Ruby 2.5+

.github/workflows/workflow.yml
```diff
@@ -14,39 +14,53 @@ jobs:
...
-          #
-          # See https://github.com/ruby/setup-ruby/issues/496
-          #
-          # This works fine on ubuntu 20.04 and 24.04.
-          # - { os: ubuntu-22.04, ruby: '2.2' }
+          # No EOL versions for windows-2022
+          - { os: macos-14, ruby: '2.5' }
+          - { os: macos-14, ruby: '2.6' }
+          - { os: macos-14, ruby: '2.7' }
+          - { os: macos-14, ruby: '3.0' }
```
MSP-Greg, line 52: Copy and paste?  Or, duplicate of the above macos-14 excludes.

nevans: Oops.  Yeah.  Thanks.  Fixed.

nevans:
> Since Ruby 2.3 & 2.4 are removed, we can change the workflow's line 46 from `uses: ruby/setup-ruby-pkgs@v1` to `uses: ruby/setup-ruby@v1`.

Thanks.  Done!

> Not sure how fast everyone wants CI to run, but adding some more Ruby 2.x jobs might be helpful.

Am I going too far with the ubuntu exclusions?  If you think so, I'll stop excluding any jobs for ubuntu-22.04 and ubuntu-24.04.

If you _really_ think we should, I'll (begrudgingly) stop excluding EOL ruby jobs for windows and macos too.  I think those are of limited utility, additional maintenance burden, _and_ they slow the build down.  But I won't argue the point any further than that.  :)

> Also, 'excluding' rather than 'including' places similar platform/os jobs in one group in the job list, which can be helpful when one is totally frustrated with CI issues...

Thanks.  Yeah I noticed that too... _after_ I pushed the changes.  Done!  I also resorted the OS list, in order to (mostly) sort the builds alphanumerically.

sodabrew:
Thanks for updating this, looks good.

PR #999, opened 2024-09-06
[CI] Windows test fixes - test_httpclient2.rb & test_pure.rb

MSP-Greg:
1. Commit fixes Windows test failures by increasing the timeout values.  These three tests also fail locally.

2. Add `# frozen_string_literal: true` to the two test files.

3. Adds a variable `@host = '127.0.0.1` and replaces in each file.  We don't need to create so many strings...

### DIFF
--- a/tests/test_httpclient2.rb
+++ b/tests/test_httpclient2.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require_relative 'em_test_helper'
 
 class TestHttpClient2 < Test::Unit::TestCase
@@ -9,6 +11,7 @@ class TestHttpClient2 < Test::Unit::TestCase
   CI_WINDOWS_OLD = windows? and RUBY_VERSION < '2.5'
 
   def setup
+    @host = '127.0.0.1'
     @port = next_port
   end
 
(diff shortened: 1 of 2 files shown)


tests/test_httpclient2.rb
```diff
@@ -107,9 +111,9 @@ def test_get_pipeline
 
   def test_authheader
     EM.run {
-      setup_timeout TIMEOUT
-      EM.start_server '127.0.0.1', @port, TestServer
-      http = silent { EM::P::HttpClient2.connect '127.0.0.1', 18842 }
+      setup_timeout(windows? ? 4 : TIMEOUT)
```
nevans, line 114: `TIMEOUT` is based on `TIMEOUT_INTERVAL`, which already a multiplier for both windows an MacOS.  Instead of supplying a straight number, what do you think about either incrementing or multiplying `TIMEOUT`?
i.e:
```suggestion
      setup_timeout(TIMEOUT + (windows? ? 2 : 0))
```
or
```suggestion
      setup_timeout(TIMEOUT * (windows? ? 1.5 : 1.0))
```

tests/test_pure.rb
```diff
@@ -51,8 +54,8 @@ def connection_completed
   def test_connrefused
     assert_nothing_raised do
       EM.run {
-        setup_timeout(2)
-        EM.connect "127.0.0.1", @port, TestConnrefused
+        setup_timeout(windows? ? 4 : 2)
```
nevans, line 57: Same question as above.  :)

nevans (approved):

tests/test_httpclient2.rb
MSP-Greg: I think I mentioned that these three tests failed on Ruby head and my Windows desktop, which is a strong system.  I just fixed the tests that failed consistently.

I think there are a lot of things in the test system that could be refactored/revised, etc.  Right now, I'm just looking to get CI passing reliably.

PR #1001, opened 2024-09-06
[CI] test files - fix Ruby head 'warning: literal string will be frozen...'

MSP-Greg:
See any Ruby head CI job's logs.

Will do another PR with lib file updates for the same.


nevans (approved):
LGTM

PR #1002, opened 2024-09-06
[CI] test_httpclient2.rb - fix intermittent failures on GHA

MSP-Greg:
Just failed in my fork.  Friday afternoon is a busy time for GHA...

Failing test makes an external connection to google.com, and sends two requests.

In Puma, we've jumped thru hoops trying to get CI stable when the GHA runners are busy.  Most issues are blocking related.  Regardless, what never fails locally often fails intermittently on GHA.

### DIFF
--- a/tests/test_httpclient2.rb
+++ b/tests/test_httpclient2.rb
@@ -88,16 +88,11 @@ class TestHttpClient2 < Test::Unit::TestCase
   def test_get_pipeline
     headers, headers2 = nil, nil
     EM.run {
-      setup_timeout TIMEOUT
+      # intermittent CI failures, external server w/two requests?
+      setup_timeout (TIMEOUT * 2)
       http = silent { EM::P::HttpClient2.connect "www.google.com", 80 }
-      d = http.get("/")
-      d.callback {
-        headers = d.headers
-      }
-      e = http.get("/")
-      e.callback {
-        headers2 = e.headers
-      }
+      http.get("/").callback { |resp| headers = resp.headers }
+      http.get("/").callback { |resp| headers2 = resp.headers }
       EM.tick_loop { EM.stop if headers && headers2 }
       EM.add_timer(1) { EM.stop }
     }


MSP-Greg:
Hold on this.  Still seeing the intermittent failures...  I'll check in my fork.

nevans:
@MSP-Greg This is probably the worst flaky test in the whole suite.  While we could improve it in several different ways (e.g: multiply TIMEOUT by 8, retry the test a couple of times on failure, fire up a local nginx in docker), EM::Protocols::HttpClient2 _does_ say in it's documentation comment: `# @deprecated Please use [EM-HTTP-Request](https://github.com/igrigorik/em-http-request) instead.`.  So, if it keeps giving us trouble, maybe we should just omit or delete this one test?

nevans:
@MSP-Greg Although... I notice that this test has something unique at the very bottom of its EM.run block.  `EM.add_timer(1) { EM.stop }` should _at least_ wait for TIMEOUT before stopping, and that's already handled by the `setup_timeout`, right?  Really, we want to stop immediately, after both callbacks succeed.  I'll make a PR for _that_.

MSP-Greg:
@nevans

After reviewing the docs, I might have a fix.  Waiting for CI in my fork to finish.

MSP-Greg:
@nevans - just pushed a fix.  All the Ubuntu jobs are passing in two runs in my fork...

nevans:
@MSP-Greg What do you think about this attempt? #1004

MSP-Greg:
@nevans

Give me a minute, but 'timeout' settings based on whether CI is running in GHA is common.   As to `test_get_pipeline` code, I think I like the following:
```ruby
http.get("/").callback { |resp| headers = resp.headers }
http.get("/").callback { |resp| headers2 = resp.headers }
```

much better than the current code:
```ruby
d = http.get("/")
d.callback {
  headers = d.headers
}
e = http.get("/")
e.callback {
  headers2 = e.headers
}
```

All the Ubuntu tests passed the test twice in my fork, but (of course) a few failed when the CI ran here.  After #999 is merged, that will help.  Maybe 'retry' should be implemented, not sure.  We use it with Puma...

tests/test_httpclient2.rb
```diff
@@ -88,16 +88,11 @@ def _test_get_multiple
...
-      d = http.get("/")
-      d.callback {
-        headers = d.headers
-      }
-      e = http.get("/")
-      e.callback {
-        headers2 = e.headers
-      }
+      http.get("/").callback { |resp| headers = resp.headers }
+      http.get("/").callback { |resp| headers2 = resp.headers }
```
sodabrew, line 95: This is definitely better style, too

MSP-Greg:
@sodabrew & @nevans

It's nice to see green CI for Ruby 2.5 thru head on all three OS's.

Next might be the OpenSSL compile warnings...

Thanks, Greg

PR #1003, opened 2024-09-06
lib files - fix Ruby head 'warning: literal string will be frozen...

MSP-Greg:
Similar to #1001, this fixes frozen string warning that are only logged with Ruby head builds.

### DIFF
--- a/lib/em/protocols/httpclient.rb
+++ b/lib/em/protocols/httpclient.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #--
 #
 # Author:: Francis Cianfrocca (gmail: blackhedd)
@@ -180,7 +182,7 @@ module EventMachine
           case @read_state
           when :base
             # Perform any per-request initialization here and don't consume any data.
-            @data = ""
+            @data = "".dup
             @headers = []
             @content_length = nil # not zero
             @content = ""
--- a/lib/em/protocols/line_protocol.rb
+++ b/lib/em/protocols/line_protocol.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module EventMachine
   module Protocols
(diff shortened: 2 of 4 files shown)


lib/em/protocols/httpclient.rb
```diff
@@ -180,7 +182,7 @@ def receive_data data
           case @read_state
           when :base
             # Perform any per-request initialization here and don't consume any data.
-            @data = ""
+            @data = "".dup
```
nevans, line 185: I haven't looked closely enough to see if it's just getting overwritten here, but the same assignment in `#post_init` (on line 92) should also be updated.  `@data` is _also_ assigned in two other places in `#receive_data`, lines 223 and 261.

Likewise, the assignment to `@content`, two lines down from here.  Curiously, `@content` is _usually_ assigned via `+=` and that probably saves it from errors, at least on the tests, but there is `<<` shovel append on line 260.

nevans (commented):
The other three files seem fine to me.  And with the suggested changes, I think `httpclient.rb` should be good to go too.

But, especially when I'm not confident of the test coverage, I've sometimes handled the conversion to frozen string literals in two PRs:
* One to add the `"".dup` changes.  This _should_ be completely safe (and removes the 3.4 chilled string warnings).
* One to add the magic comments.  This is where the gaps in test coverage will bite us in production. 🙂

nevans:
I forgot this was already merged.  I'll just submit a follow-up PR.  🙂

lib/em/protocols/httpclient.rb
nevans: Ah... the "two other places in `#receive_data` are both the lvar `data`, not `@data`.  The lvar isn't being mutated, so those seem fine.

PR #1006, opened 2024-09-11
Run the full test suite in pure ruby CI

nevans:
~Many of these may be trivial to implement in pure ruby mode, and some may be impossible, but for right now I'd prefer to simply have them all displayed as a "TODO list" in the test results.~

~Of course, adding these stub methods breaks many tests that were omitted based on whether or not `EM` responds to these methods.~

All of the failing tests have been marked as pending, _including the ones that failed when stub methods were added and raised EM::Unsupported_.  Where the tests failed due to an EM::Unsupported error, the name of the stubbed method has been added to the pending message.  This should simplify finding tests to un-pend, if and when pure ruby mode is fixed to support these features.

Additionally, CI for pure ruby mode has been updated to run _the entire_ test suite--albeit with 90 tests marked as pending.  This should allow us to "ratchet" pure ruby mode forward.

~The major downside~ ~A minor tradeoff of this PR is that `EM.respond_to?(method)` is no longer a reliable feature detection mechanism.  (see the comment below for an analysis of this issue)~ _This has been moved into another PR._

### DIFF
--- a/lib/em/pure_ruby.rb
+++ b/lib/em/pure_ruby.rb
@@ -242,6 +242,26 @@ module EventMachine
     def epoll
     end
 
+    # Pure ruby mode does not allow setting epoll
+    # @private
+    def epoll=(bool)
+      bool and raise Unsupported, "EM.epoll is not supported in pure_ruby mode"
+    end
+
+    # Pure ruby mode does not support epoll
+    # @private
+    def epoll?;  false end
+
+    # Pure ruby mode does not support kqueue
+    # @private
+    def kqueue?; false end
+
+    # Pure ruby mode does not allow setting kqueue
+    # @private
+    def kqueue=(bool)
+      bool and raise Unsupported, "EM.kqueue is not supported in pure_ruby mode"
+    end
+
     # @private
     def ssl?
       true
--- a/rakelib/test_pure.rake
+++ b/rakelib/test_pure.rake
@@ -7,7 +7,7 @@ end
 
 task :test_em_pure_ruby do
   ENV['EM_PURE_RUBY'] = 'true'
-  Rake::Task['test_pure'].execute
+  Rake::Task['test'].execute
 end
 
 task test_em_pure_ruby: "test:fixtures"
--- a/rakelib/test_pure.rake
+++ b/rakelib/test_pure.rake
@@ -1,10 +1,5 @@
 require 'rake/testtask'
 
-Rake::TestTask.new(:test_pure) do |t|
-  t.test_files = Dir.glob('tests/**/test_pure*.rb') + Dir.glob('tests/**/test_ssl*.rb')
-  t.warning = true
-end
-
 task :test_em_pure_ruby do
   ENV['EM_PURE_RUBY'] = 'true'
   Rake::Task['test'].execute
--- a/tests/test_attach.rb
+++ b/tests/test_attach.rb
@@ -44,6 +44,7 @@ class TestAttach < Test::Unit::TestCase
   end
 
   def test_attach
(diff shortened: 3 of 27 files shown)


sodabrew:
This look fantastic. I'd like to merge it as-is, just note if you think you're still working on it or ready to land!

sodabrew:
... thinking some more:
> The major downside of this PR is that EM.respond_to?(method) is no longer a reliable feature detection mechanism.
I'm using GitHub and GitLab code search to see if this is a common pattern, I did find a few:
https://github.com/search?q=%22%20EM.respond_to%3F%22++path%3A*.rb&type=Code&ref=advsearch&l=&l=

nevans:
> ... thinking some more:
> 
> > The major downside of this PR is that EM.respond_to?(method) is no longer a reliable feature detection mechanism.
> 
> I'm using GitHub and GitLab code search to see if this is a common pattern, I did find a few: https://github.com/search?q=%22%20EM.respond_to%3F%22++path%3A*.rb&type=Code&ref=advsearch&l=&l=

Thanks for taking that next step.  Yeah, this is the potential issue.  I looked at all of the examples, and also searched for `EventMachine.respond_to?`, and I think we're okay.  I'm fairly certain _all_ of the search results are one of the following:
* Many results are simply from eventmachine/eventmachine or forks: they are simply the `respond_to?` calls in our own test suite.
* Several repos have vendored copies of eventmachine, so the same deal there as the forks.
* Most of the (not forked or bundled) repos were checking for a "public" (and documented as public) method that isn't stubbed or changed by this PR.  These are basically using `EM.respond_to?` to feature test for the _version_ of EventMachine, not for the implementation mode.  For example: `watch`, `reactor_running?`, `reactor_thread`, `attach`, `defers_finished?`, `tick_loop`, 
* In at least one repo, they were guarding against their own `alias` method chaining (thus unchanged by this PR).
* In at least one repo, they were checking for something added by a different gem (thus unchanged by this PR).
* In at least one repo, it was `em.respond_to?` for an abbreviated "embed" argument (i.e: not EventMachine).

But there's a deeper reason I think we're okay: pretty close to _no one_ should be using pure ruby mode today... or if they are, they're dealing with lots and lots of bugs already, and that's regardless of whether or not they are using `respond_to?` for feature detection.  Because there are still some basic pieces of broken functionality in pure ruby mode.  :)  This PR only really breaks `respond_to?` if you enable pure ruby mode, but that approach is already broken.  :)

Additionally, most of the stubbed methods are documented as `@private`, and many of them require a `signature` parameter or return a signature.  Likewise, most of them have another related method that is documented and doesn't work with event descriptor signatures.  So, although I wish these quasi-private methods were handled differently, it looks like users of EventMachine are mostly ignoring them and not making dependencies on them.  That's good.

nevans:
@sodabrew I still think it's safe for the reasons given, but its not completely risk-free, so I'm also fine with putting that part off for a future PR.  I've removed almost all of the stubs.  The only stub methods this PR adds now are the ones related to epoll and kqueue, so  `rake test` doesn't immediately crash.

There are still many files that check `EM.respond_to?(method)` before even defining certain test case classes or methods, and those won't be added to the "pending" totals.  Even without those, this is still a huge improvement.  Compare to the numbers in https://github.com/eventmachine/eventmachine/pull/1006#issuecomment-2343945873:

```
Finished in 31.088036539 seconds.
-------------------------------------------------------------------------------
301 tests, 3482 assertions, 0 failures, 0 errors, 76 pendings, 14 omissions, 0 notifications
73.5192% passed
-------------------------------------------------------------------------------
9.68 tests/s, 112.00 assertions/s
```

PR #1015, opened 2025-02-12
Implement pure_ruby `EM.current time`

nevans:
This already existed on `EM::Reactor#current_loop_time`, we just needed to forward to it from the singleton method on `EM`.  This enables four more existing tests.  They are very slow in pure ruby mode, but they do pass!

I converted the current_loop_time to use `CLOCK_MONOTONIC`, for the obvious reasons.

And current_loop_time is now stored as a microsecond integer (not a `Time` object):
* Integer doesn't use ObjectSpace or malloc (prior to variable width allocation for Time).
  The reactor loop is the kind of hotspot where that's worth considering.
* This matches the behavior of the C++ extension's `EventMachine_t::GetRealTime()`.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -25,7 +25,7 @@ jobs:
         os:
           - ubuntu-22.04
           - ubuntu-24.04
-          - macos-14 # arm64
+          - macos-15 # arm64
           - windows-2022
 
         ruby: ${{ fromJson(needs.ruby-versions.outputs.versions) }}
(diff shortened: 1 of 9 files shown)


nevans:
n.b: this includes #1018, so CI can pass.

bahmn1 (approved):

sodabrew:
@copilot update to latest main

nevans:
I rebased this on top of #1028, so the tests can pass for modern ruby versions.

nevans:
Note: I've rebased this on #1028 and #1029, so CI should be able to pass now.

PR #1017, opened 2025-02-12
Fix test fixture CSR generation

nevans:
See ruby/openssl#734 and ruby/openssl#747.


nevans:
This was merged as a part of #1018.

PR #1016, opened 2025-02-12
Add `ostruct` to the Gemfile (for ruby 3.5)

nevans:

PR #1018, opened 2025-02-12
Update CI for changes to ruby 3.5 and openssl

nevans:
I originally created a different PR and then a second one when I thought that getting the build passing again would be only one or two things.  😉  So this PR combines:
* #1017 
* #1016 
* b5cd412 - **Dynamically update new ruby versions**
  This is the same ruby versions action used by all stdlib gems, so it will keep CI up-to-date with new ruby releases automatically.
* f34b077 - **Skip `send_data_for` tests for Windows with Ruby 3.4**
  I don't know what's going on with this. 🙁  But it seems to fail consistently (not flaky).  It works fine with ruby 3.5-dev, so maybe it's a ruby 3.4.1 bug?
* 09e3f7c - **Run pure_ruby tests in same job as regular tests**
  I'm just playing the odds here: some random percent of the time, one of the jobs just behaves poorly and causes the flaky tests to fail.  So, the more jobs in the matrix, the more likely the build fails.  So, reducing the jobs in the matrix should allow the build to successfully complete more often.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -3,7 +3,16 @@ name: CI
 on: [ push, pull_request, workflow_dispatch ]
 
 jobs:
+  ruby-versions:
+    uses: ruby/actions/.github/workflows/ruby_versions.yml@master
+    with:
+      engine: cruby
+      # We currently support four EOL versions of ruby, but only on linux
+      # versions that were released before that version of ruby went EOL.
+      min_version: 2.5
+
   build:
+    needs: ruby-versions
     name: >-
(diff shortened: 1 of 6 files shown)


nevans:
@sodabrew This _does_ pass... it's just still a little bit flaky.  Not sure why, some github actions workers are just randomly much slower than others, and I think that may be what's causing the issues.

nevans:
@sodabrew although 09e3f7c still had one flaky job randomly fail in this PR's checks, it allowed #1015, #1019, and #1020 to succeed.  3 out of 4 is better odds than I'd been seeing. 😉

sodabrew:
Nice work, thank you!

Issue #1014, opened 2025-01-13
NetBSD 10.1 evbarm (RPI) - unable to compile eventmachine-1.2.7


drixter:
Hello Team,

Below the result to compile eventmachine-1.2.7 on NetBSD 10.1 evbarm (RPI)
It's faling, any hint to check something more?

`gem32 install jekyll bundler eventmachine`

```
[NetBSD-drixter@heaven:~]$ gcc --version
gcc (nb3 20231008) 10.5.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
[NetBSD-drixter@heaven:~]$ gem32 --version
3.4.19
[NetBSD-drixter@heaven:~]$ ruby32 --version
ruby 3.2.6 (2024-10-30 revision 63aeb018eb) [aarch64-netbsd]
```

```
current directory: /home/users/drixter/gems/gems/eventmachine-1.2.7/ext
/usr/pkg/bin/ruby32 extconf.rb
checking for -lcrypto... yes
checking for -lssl... yes
checking for openssl/ssl.h... yes
checking for openssl/err.h... yes
checking for rb_trap_immediate in ruby.h,rubysig.h... no
checking for rb_thread_blocking_region()... no
checking for rb_thread_call_without_gvl() in ruby/thread.h... yes
checking for rb_thread_fd_select()... yes
checking for rb_fdset_t in ruby/intern.h... yes
checking for rb_wait_for_single_fd()... yes
checking for rb_enable_interrupt()... no
checking for rb_time_new()... yes
checking for inotify_init() in sys/inotify.h... no
checking for __NR_inotify_init in sys/syscall.h... no
checking for writev() in sys/uio.h... yes
checking for pipe2() in unistd.h... yes
checking for accept4() in sys/socket.h... yes
checking for SOCK_CLOEXEC in sys/socket.h... yes
checking for sys/event.h... yes
checking for sys/queue.h... yes
checking for clock_gettime()... yes
checking for CLOCK_MONOTONIC_RAW in time.h... no
checking for CLOCK_MONOTONIC in time.h... yes
CXXFLAGS=-O2 -pthread -I/usr/pkg/include -I/usr/include -Wall -Wextra -Wno-deprecated-declarations -Wno-ignored-qualifiers -Wno-unused-result -Wno-address
creating Makefile

[...]

PR #1020, opened 2025-02-14
Apply NetBSD pkgsrc patch

nevans:
At @drixter's suggestion (#1014), I've copied the patch from NetBSD's pkgsrc, commit [d85ec4a9d](https://github.com/NetBSD/pkgsrc/commit/d85ec4a9dd39ef319a6e71ee67e3eb33ef8b6a10)
[devel/ruby-eventmachine/patches/patch-ext_em.cpp](https://github.com/NetBSD/pkgsrc/blob/trunk/devel/ruby-eventmachine/patches/patch-ext_em.cpp).

This is based on #1018 _(rebased on 2026-07-07 on #1028, #1029)_, so the build will pass (maybe?).

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -3,7 +3,16 @@ name: CI
 on: [ push, pull_request, workflow_dispatch ]
 
 jobs:
+  ruby-versions:
+    uses: ruby/actions/.github/workflows/ruby_versions.yml@master
+    with:
+      engine: cruby
+      # We currently support four EOL versions of ruby, but only on linux
+      # versions that were released before that version of ruby went EOL.
+      min_version: 2.5
+
   build:
+    needs: ruby-versions
     name: >-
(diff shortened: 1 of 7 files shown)


StantonMatt (commented):
I checked the C++ hunk against the NetBSD pkgsrc patch for ruby-eventmachine, and it matches the intended kqueue guard: keep `<sys/param.h>` for `__NetBSD__` and only use the `NOTE_EOF` fflags form before NetBSD 9.99.15.

On macOS arm64 with Ruby 3.3.11, this branch compiled cleanly for me with `bundle exec rake compile`, and `bundle exec ruby -Ilib -Itests tests/test_send_file.rb` passed with 7 tests / 7 assertions. I can't exercise the NetBSD path from here, and the PR currently reports no hosted checks, but the local compile/send_file smoke looks good from my side.

nevans:
Note: I've rebased this on #1028 and #1029, so CI should be able to pass now.

PR #1024, opened 2026-06-02
Use typed data for fast file reader mapper

StantonMatt:
The fast file reader extension still used `Data_Wrap_Struct` and `Data_Get_Struct` for `EventMachine::FastFileReader::Mapper`. Ruby 4.1-dev no longer exposes those legacy macros to this C++ extension path, so downstream Ruby-head installs fail while compiling `ext/fastfilereader/rubymain.cpp`.

This switches the mapper wrapper/accessors to Ruby typed-data APIs while keeping the same free function and object lifetime.

Verification:

- `bundle exec rake compile` with Homebrew Ruby 3.3.11 on arm64-darwin25
- `bundle exec ruby -Ilib -Itests tests/test_send_file.rb` -> 7 tests, 7 assertions
- direct `EventMachine::FastFileReader::Mapper` smoke for `size`, `get_chunk`, and `close`
- isolated fast file reader extension compile with system Ruby 2.6 headers
- `git diff --check`
- clean review-fix-loop pass

I also tried the full `bundle exec rake test` locally, but it stops before running tests while generating SSL fixtures under Ruby 3.3/OpenSSL 3.6 (`X509_REQ_set_version: passed invalid argument`). I left that out of scope because the focused fast-file-reader test path passes and the failure happens before this extension code is exercised.

### DIFF
--- a/ext/fastfilereader/rubymain.cpp
+++ b/ext/fastfilereader/rubymain.cpp
@@ -41,6 +41,12 @@ static void mapper_dt (void *ptr)
 		delete (Mapper_t*) ptr;
 }
 
+static const rb_data_type_t mapper_type = {
+	"EventMachine::FastFileReader::Mapper",
+	{0, mapper_dt, 0,},
+	0, 0, RUBY_TYPED_FREE_IMMEDIATELY
+};
+
 /**********
 mapper_new
 **********/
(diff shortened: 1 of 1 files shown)


nevans:
This can be closed now.  It was also done by #1027, which was merged.

PR #1026, opened 2026-06-23
Skip SSLv3 tests on OpenSSL 4.0+

raviksharma:
SSLv3 handshake attempts fail with "no protocols available" in OpenSSL 4.0,
even when OPENSSL_NO_SSL3 is not defined at compile time. The existing
omit guard only checks the compile-time flag.

Add IS_SSL_GE_4_0 runtime version check (following the existing
IS_SSL_GE_1_1 pattern) and skip the 4 SSLv3 tests when running on
OpenSSL 4.0+.

Co-authored-by: Hermes Agent (z-ai/glm-5.2) <noreply@nousresearch.com>


sodabrew:
Thank you!

PR #1027, opened 2026-07-06
Migrate fastfilereader to the TypedData API

yahonda:
## Problem

Ruby master removed the long-deprecated `Data_Wrap_Struct` / `Data_Get_Struct` macros in [ruby/ruby@a26f528b3b](https://github.com/ruby/ruby/commit/a26f528b3b) ("Deprecate `struct RData` in favor of `struct RTypedData`"), so the fastfilereader extension no longer compiles against ruby-head (4.1.0dev):

```
rubymain.cpp:53:19: error: 'Data_Wrap_Struct' was not declared in this scope; did you mean 'TypedData_Wrap_Struct'?
```

This makes `gem install eventmachine` fail on ruby-head, which also breaks CI of downstream projects that bundle eventmachine — e.g. mysql2's ruby-head job: https://github.com/brianmario/mysql2/actions/runs/28774101830/job/85314025661

## Fix

Wrap `Mapper_t` with `TypedData_Wrap_Struct` / `TypedData_Get_Struct` and an `rb_data_type_t` instead. The TypedData API predates the gem's `required_ruby_version` (>= 2.5.0), so no compatibility is lost.

## Verification

- `rake compile` succeeds on ruby 4.0.5 and on ruby master (4.1.0dev, `96311d13c2`); previously the latter failed
- Full test suite on ruby master: 315 tests, 3657 assertions, 0 failures (the two `TestIPv6` errors on my machine are environmental; `test_send_file` is no longer skipped)
- `gem build` + `gem install` of the built gem succeeds on ruby master and `require "fastfilereaderext"` loads
- mysql2's spec suite (including its EventMachine specs) passes on ruby master when bundling this branch: 352 examples, 0 failures

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ext/fastfilereader/rubymain.cpp
+++ b/ext/fastfilereader/rubymain.cpp
@@ -41,6 +41,12 @@ static void mapper_dt (void *ptr)
 		delete (Mapper_t*) ptr;
 }
 
+static const rb_data_type_t mapper_type = {
+	"EventMachine::FastFileReader::Mapper",
+	{ 0, mapper_dt, 0, },
+	0, 0, RUBY_TYPED_FREE_IMMEDIATELY
+};
+
 /**********
 mapper_new
 **********/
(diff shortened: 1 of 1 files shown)


sodabrew:
Oh wow, the TypeData macros have been available since Ruby 1.9.2 according to https://bugs.ruby-lang.org/issues/19998

PR #1029, opened 2026-07-07
Drop ruby 2.5 support (bump minimum ruby to 2.6.0)

nevans:
As per the policies adopted in #998:
* I've dropped all EOL ruby versions from windows and macos runners.
* I've dropped all OS versions that GitHub has stopped supporting.  They simply don't run anyway.
* I've upgraded the macos and windows runners to their "latest" GA versions.
* Since GitHub dropped support for Ubuntu versions that were released prior to ruby 2.6.0, we aren't running CI against ruby 2.5.

Therefore, since we aren't running CI against ruby 2.5, we should bump the minimum required ruby version to 2.6.

Note that this PR includes #1028, so the `pure_ruby` tests also pass.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -25,7 +25,7 @@ jobs:
         os:
           - ubuntu-22.04
           - ubuntu-24.04
-          - macos-14 # arm64
+          - macos-15 # arm64
           - windows-2022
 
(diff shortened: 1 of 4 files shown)


nevans:
@sodabrew Thanks for merging those other PRs yesterday.  This PR fixes the failing builds (there are still some flaky tests, but...)

